PR #2, opened 2022-02-26
Cora code review

StuartCowley:


src/strings.js
```diff
@@ -1,27 +1,34 @@
...
   // your code here
+  return string.charAt(0);
 };
 
 function firstCharacters (string, n) {
   // your code here
+  return string.substr(0, n);
 };
 
+
```
StuartCowley, line 31: Blank line isn't needed and can be removed
```suggestion
```

src/strings.js
```diff
@@ -1,27 +1,34 @@
 function sayHello (string) {
   // your code here
+  return "Hello, " + string + "!";
```
StuartCowley, line 3: All the `// your code here` comments should be removed

src/objects.js
```diff
@@ -1,43 +1,67 @@
...
 };
 
 const createTalkingPerson = (name, age) => {
-  // your code here
+  return {
+    name: name,
+    age: age,
+    introduce: (firendsName) => {
+      return `Hi ${firendsName}, my name is ${name} and I am ${age}!`;
+    }
```
StuartCowley, line 58: Just a typo
```suggestion
    introduce: (friendsName) => {
      return `Hi ${friendsName}, my name is ${name} and I am ${age}!`;
    }
```

src/objects.js
```diff
@@ -1,43 +1,67 @@
...
+  }
 };
 
+
+
+
```
StuartCowley, line 64: These three lines arent needed, should be removed
```suggestion
```

src/objects.js
```diff
@@ -1,43 +1,67 @@
 const createPerson = (name, age) => {
   // your code here
+  return {
```
StuartCowley, line 3: All the `// your code here` comments should be removed

src/arrays.js
```diff
@@ -1,61 +1,126 @@
 const getNthElement = (index, array) => {
   // your code here
+  if (index > array.length - 1) {
+    index -= array.length;
+  }
+  return array[index];
```
StuartCowley, line 6: As before the `// your code here` comment should be removed

This could be simplified using the remainder operator:
```suggestion
  return array[index % array.length];
```

src/arrays.js
```diff
@@ -1,61 +1,126 @@
...
 };
 
 const arrayToCSVString = array => {
   // your code here
+  return array.reduce((total, element) => `${total},${element}`);
```
StuartCowley, line 11: Interesting solution!

src/arrays.js
```diff
@@ -1,61 +1,126 @@
...
 };
 
 const csvStringToArray = string => {
   // your code here
+  const newArr = [];
```
StuartCowley, line 16: This new array isnt being used, should be deleted

src/arrays.js
```diff
@@ -1,61 +1,126 @@
...
+  );
+  // 4. map through the new array and reduce each array of characters to a string (you get the reversed words into a new array)
+  const finalArrayOfReversedWords = arrayOfArraysOfReversedCharacters.map(
+    array => {
+      return array.reduce((reversedWord, currentLetter) => {
+        return reversedWord + currentLetter;
+      });
+    }
+  );
+  return finalArrayOfReversedWords;
```
StuartCowley, line 67: Theres a lot of noise here in the comments that doesnt need to be. It looks like you're trying to explain _how_ you're doing something rather than _why_, which is more valuable for whoever will come to read this in the future. As these are standard methods it would be easy enough to look up what they are doing if the reader is unsure, so these comments should all be removed to make it easier to read.
While this works, it is a bit over complicated, and as it happens in this case the logic can be greatly simplified using almost the same methods you have but with a bit of refactoring:

```suggestion
  return strings.map(string =>
    string
      .split('')
      .reverse()
      .join('')
  );
```

src/arrays.js
```diff
@@ -1,61 +1,126 @@
...
   // your code here
+  if (index === undefined) {
+    return array;
+  }
+  const newArr = [];
+  for (let i = 0; i < array.length; i += 1) {
+    newArr[i] = array[i];
+  }
+  newArr.splice(index, 1);
+  return newArr;
```
StuartCowley, line 91: Theres a couple of things here:
- The `if` block on line 83 isnt needed at all,  removing it didnt cause any failing tests so it wasnt adding any value,
- The use of a for loop to iterate over the existing array and add its values directly to a new one without any logic to change its values (which isnt needed in this case) could be removed and instead a spread operator used to simply copy the values into the new array without the extra work:
```suggestion
  const newArr = [...array];
  newArr.splice(index, 1);
  return newArr;
```
This does work, but this just makes the process more efficient and cleaner

src/arrays.js
```diff
@@ -1,61 +1,126 @@
...
+      returnArr.push(strings[i]);
+    }
+  }
+  return returnArr;
 };
 
 const removeSpaces = string => {
   // your code here
+  const array = string.split(' ');
+  return array.join('');
```
StuartCowley, line 108: Could be simplified a bit to use the same methods on one line without the variable creation:
```suggestion
  return string.split(' ').join('');
```

src/booleans.js
```diff
@@ -1,62 +1,100 @@
...
+}
 
 function truthiness(a) {
   // your code here
-};
+  if (a === 0 || Number.isNaN(a) || a === null || a === undefined || a === '') {
+    return false;
+  }
+
+  return true;
```
StuartCowley, line 32: This is quite overengineered, you dont need to supply an exhaustive list of things that are falsy when its built into the language, have a look at https://developer.mozilla.org/en-US/docs/Glossary/Falsy for examples.
A cleaner way is to simply pass the value into the Boolean constructor and return that:
```suggestion
  return Boolean(a);
```

src/booleans.js
```diff
@@ -1,62 +1,100 @@
...
+      if (string[i].toLowerCase() === vowels[j]) {
+        vowelCount += 1;
+      }
+    }
+  }
+  if (vowelCount > 0) {
+    return true;
+  }
+  return false;
+}
```
StuartCowley, line 85: Interesting solution! While it works, the root of what the function is trying to do here is pattern recognition of the strings, in which case a regex could be a good option (as with the remove element from array solution you did). This would mean you dont have to run two for loops inside each other and can simply check if the string has any elements that match what you provide, in this case, upper and lower case vowels:
`return /[aeiou]/i.test(string);`
The characters in the brackets are what is being checked for, and the `/i` at the end stands for case-insensitive so thats also accounted for

I wouldnt expect you to have known that off hand this early on though so don't worry! Regex is one of those things that will keep coming up and it gets easier with time

src/booleans.js
```diff
@@ -1,62 +1,100 @@
...
   // your code here
-};
+  const up = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+  for(let i = 0; i < string.length; i += 1) {
+      for(let j = 0; j < up.length; j += 1) {
+          if(string[i] === up[j]) {
+              return false;
+          }
+      } 
+  } return true;
```
StuartCowley, line 96: Theres a built in function for making a string lower case as you've used above, how about making the string lower case using that and then comparing it to the original? If it matches then it was already lower case

src/numbers.js
```diff
@@ -1,45 +1,59 @@
...
   // your code here
+  return Math.abs(a);
 }
 
 function quotient (a, b) {
   // your code here
+  if(a < 0 || b < 0) {
+    return Math.ceil(a / b);
+  }
+  return Math.floor(a / b);
```
StuartCowley, line 51: Looking at what a quotient is as defined in the test:
```
  // N.B. quotient of two numbers is the integer given by dividing
  // the first by the second, without the remainder
  // 18 divided by 7 is 2 remainder 4 (or 2.571...)
  // so the quotient of 18 and 7 is 2
```
What we are looking to do is less about rounding numbers up or down, but simply removing anything after the decimal point. There is a built in function in JS to do just that (`Math.trunc()`) which will simplify this a bit:

```suggestion
  return Math.trunc(a / b);
```

StuartCowley (commented):
Some good work here and some interesting approaches! Theres a few suggestions added about comments, cleaning up a few things, and. some refactoring suggestions that make use of existing javascript functions that will make the code more concise and save you having to write customer solutions, Ive included some examples of these and given a few hints about where to find the answers to some of the others

Overall good effort, everything passes!

PR #3, opened 2022-02-27
Ash code review

StuartCowley:


src/booleans.js
```diff
@@ -1,61 +1,65 @@
-function negate(a) {
-  // your code here
+const negate = (a) => {
+  return !(a);
```
StuartCowley, line 2: You dont need these brackets here, this works just as well:
```suggestion
  return !a;
```

This is the case for most of the return statements in this file

src/arrays.js
```diff
@@ -1,61 +1,71 @@
...
 };
 
 const reverseWordsInArray = strings => {
-  // your code here
+  return strings.map(name => name.split('').reverse().join(''));
 };
 
 const onlyEven = numbers => {
-  // your code here
+  return numbers.filter(item => item % 2 == 0);
```
StuartCowley, line 42: Would be better to use the strict equals here:
```suggestion
  return numbers.filter(item => item % 2 === 0);
```

src/arrays.js
```diff
@@ -1,61 +1,71 @@
...
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  return strings.filter(string => /^[aeiouAEIOU]/i.test(string));
 };
 
 const removeSpaces = string => {
-  // your code here
+  string = string.replace(/\s+/g, '');
+  return string;
```
StuartCowley, line 57: Could just return it directly, you don't need to create the extra variable here:
```suggestion
  return string.replace(/\s+/g, '');
```

src/arrays.js
```diff
@@ -1,61 +1,71 @@
...
 const removeNthElement2 = (index, array) => {
-  // your code here
+  const newArray = [...array];
+  newArray.splice(index, 1);
+  return newArray;
 };
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  return strings.filter(string => /^[aeiouAEIOU]/i.test(string));
```
StuartCowley, line 52: The `/i` flag at the end of the regex here stands for case-insensitive, so it will match both cases:
```suggestion
  return strings.filter(string => /^[aeiou]/i.test(string));
```

src/arrays.js
```diff
@@ -1,61 +1,71 @@
 const getNthElement = (index, array) => {
-  // your code here
+  if (index > array.length - 1) {
+    return (array[index - array.length]);
+  } else {
+  return array[index];
+  }
```
StuartCowley, line 6: This could be done much more simply with the remainder operator:
```suggestion
  return array[index % array.length];
```

src/arrays.js
```diff
@@ -1,61 +1,71 @@
...
+  string = string.replace(/\s+/g, '');
+  return string;
 };
 
 const sumNumbers = numbers => {
-  // your code here
+  let sum = 0;
+for (let i = 0; i < numbers.length; i++) {
+    sum += numbers[i];}
+  return sum;
```
StuartCowley, line 64: While this works, there is a cleaner function built into JS that will do this just fine:
```suggestion
  return numbers.reduce((prev, current) => prev + current, 0);
```

src/booleans.js
```diff
@@ -1,61 +1,65 @@
...
 };
 
-function truthiness(a) {
-  // your code here
+const truthiness = (a) => {
+  if (a) {
+    return true
+  } else {
+    return false
+  }
```
StuartCowley, line 26: This works, but there is a cleaner way. Remember all values in JS are either truthy or falsy:
'dfdf' = truthy
true = true
0 = false
undefined = false

So we cannot just return. whatever is passed in, but what we can do is negate it to turn it into a boolean value with the `!` not operator: 
if a = 'dfdf' (truthy), then !a == false
true = true (truthy), then !a == false
0 = false (falsy), then !a == true

So by negating _that_ value we can work out the truthiness of the original value, eg
if a = 'dfdf' (truthy), then !a == false so !!a == true

So for this function we could just return:
```suggestion
  return !!a;
```

Hope that makes sense!

src/booleans.js
```diff
@@ -1,61 +1,65 @@
...
-function startsWith(char, string) {
-  // your code here
+const startsWith = (char, string) => {
+  return string.startsWith(char);
 };
 
-function containsVowels(string) {
-  // your code here
+const containsVowels = (string) => {
+  return string.match('a|e|i|o|u|A|E|I|O|U/i') ? true : false;
```
StuartCowley, line 58: A couple of things with this one:
- the case-insensitive `/i` flag at the end means you dont have to put in both upper and lower case letters to get a match,
- The ternary isnt needed, as the function should return a truthy value so this is just adding a bit of extra noise,
- Theres a very similar solution to this as what you used in the arrays file:
```suggestion
  return /[aeiou]/i.test(string);
```

I wouldnt worry too much about memorising the details about how regex works this early on though. Its one of those things that will get easier the more you do it and become more familiar with the use cases and weird syntax

src/booleans.js
```diff
@@ -1,61 +1,65 @@
...
 };
 
-function isLowerCase(string) {
-  // your code here
+const isLowerCase = (string) => {
+  return string == string.toLowerCase() && string != string.toUpperCase();
```
StuartCowley, line 62: This comparison doesnt make a lot of sense, you are comparing if the string is all lower case with the first statement, and if the string is not all uppercase with the second one. This means this will wrongly pass if a test is given for an all capital string.
Instead you could just test if the string is all lower case as follows:
```suggestion
  return string === string.toLowerCase();
```

src/objects.js
```diff
@@ -1,41 +1,55 @@
...
+  return object.hasOwnProperty(property);
 };
 
 const isOver65 = person => {
-  // your code here
+  if (person.age > 65) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 26: This works but could be simplified quite a bit:
```suggestion
  return person.age > 65;
```

src/objects.js
```diff
@@ -1,41 +1,55 @@
...
 const averageAge = people => {
-  // your code here
+  return people.reduce((acc, curr) => acc + curr.age, 0) / people.length;
 };
 
 const createTalkingPerson = (name, age) => {
-  // your code here
+  return {
+    name: name,
+    age: age,
```
StuartCowley, line 48: Minor point, but there is a shorthand for this that shortens the code a little bit:
```suggestion
    name,
    age,
```

StuartCowley (commented):
Good work, all passes and for the most part nice clean code. Nice use of ES6 arrow function syntax throughout too! Left a few suggestions about how to tidy up a few parts here and there as well as a few explanations to help out a bit, hope its useful

PR #4, opened 2022-02-27
Vee code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,84 @@
 const getNthElement = (index, array) => {
-  // your code here
+  // if (index === array.length
+  // return array.at(index);
+
+  // if (index > array.length - 1) {
+  //   index -= array.length;
+  // }
+  // return array[index];
+
```
StuartCowley, line 9: Can delete all this, just adding clutter

src/arrays.js
```diff
@@ -1,61 +1,84 @@
...
 };
 
 const arrayToCSVString = array => {
-  // your code here
+  return array.join(',');
 };
 
 const csvStringToArray = string => {
-  // your code here
+  return Array.from(string.replace(/,/g, ''));
```
StuartCowley, line 18: Interesting solution!

src/arrays.js
```diff
@@ -1,61 +1,84 @@
...
 };
 
 const numbersToStrings = numbers => {
-  // your code here
+  return numbers.toString().split(',');
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+  return strings.map(strings2 => strings2.toUpperCase());
```
StuartCowley, line 38: Minor point, with map its better to think about the temporary variable you create as a singular instance of what the array contains, so a string in this case:
```suggestion
  return strings.map(string => string.toUpperCase());
```

src/arrays.js
```diff
@@ -1,61 +1,84 @@
...
 
 const reverseWordsInArray = strings => {
-  // your code here
+  return strings
+    .toString()
+    .split('')
+    .reverse()
+    .join('')
+    .split(',')
+    .reverse();
```
StuartCowley, line 48: Interesting approach here. I see what you were thinking, but there is a cleaner way to do this using the same methods:
```suggestion
  return strings.map(string => {
    return string
      .split('')
      .reverse()
      .join('');
  });
```

src/arrays.js
```diff
@@ -1,61 +1,84 @@
...
+  return array.slice(0, index).concat(array.slice(index + 1));
 };
 
-const elementsStartingWithAVowel = strings => {
-  // your code here
-};
+function elementsStartingWithAVowel(strings) {
+  // strings.filter(/^[aeiou]/i.test(strings));
+  return strings.filter(e => /^[aeiou]/i.test(e));
+  // /[aeiou]/i.test(string);
```
StuartCowley, line 64: Comments can be removed here
```suggestion
  return strings.filter(e => /^[aeiou]/i.test(e));
```

src/arrays.js
```diff
@@ -1,61 +1,84 @@
...
+}
 
 const removeSpaces = string => {
-  // your code here
+  console.log(string.replace(/\s/g, ''));
```
StuartCowley, line 68: `Console.log`s should be removed before submitting code for review unless its part of required functionality

src/arrays.js
```diff
@@ -1,61 +1,84 @@
...
+    .join('')
+    .split(',')
+    .reverse();
 };
 
 const onlyEven = numbers => {
-  // your code here
+  return numbers.filter(function(num) {
+    return num % 2 === 0;
+  });
```
StuartCowley, line 54: You dont need to add the `function` in here like this, can use the arrow syntax:
```suggestion
  return numbers.filter(number => number % 2 === 0);
```

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
 
 function both(a, b) {
-  // your code here
-};
+  //   if (a === true && b === true) {
+  //     return true;
+  //   }
+  //   return false;
+  // }
+  return a === true && b === true;
```
StuartCowley, line 11: You dont need to explicitly check truthiness like this, remember all values in JS are either truthy or falsy so you can just compare them directly in cases like this:
```suggestion
  return a && b;
```

Also the comments should be removed

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+}
 
 function either(a, b) {
-  // your code here
-};
+  if (a === true || b === true) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 19: As above, can be greatly simplified:
```suggestion
  return a || b;
```

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+}
 
 function none(a, b) {
-  // your code here
-};
+  if (a === false && b === false) {
```
StuartCowley, line 23: See comments above

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 function one(a, b) {
-  // your code here
-};
+  if (a === true && b === false) {
```
StuartCowley, line 31: See comments above

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
 
 function isOdd(a) {
-  // your code here
-};
+  let remainder = a % 2;
+  if (remainder === 0) {
+    return false;
+  } else {
+    return true;
+  }
```
StuartCowley, line 74: The logic here is sound, but a bit long winded. All the necessary parts are there, but it can be simplified a bit!
```suggestion
  return a % 2 !== 0;
```

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+}
 
 function isEven(a) {
-  // your code here
-};
+  let remainder = a % 2;
```
StuartCowley, line 78: Same applies as in the `isOdd` function

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
-};
+  // if (string.toLowerCase() === string) {
+  //   return true;
+  // } else {
+  //   return false;
+  // }
+  //   let a;
+  //   string.toLowerCase() === string ? (a = true) : (a = false);
+  //   return a;
+  // }
```
StuartCowley, line 117: Can remove all these comments

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 function isGreaterThan(a, b) {
-  // your code here
-};
+  if (a > b) {
```
StuartCowley, line 53: See above about simplicity of code

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 function isLessThanOrEqualTo(a, b) {
-  // your code here
-};
+  if (a <= b) {
```
StuartCowley, line 61: See above about simplicity of code

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 function startsWith(char, string) {
-  // your code here
-};
+  if (string[0] === char) {
```
StuartCowley, line 95: See above about simplicity of code

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 function containsVowels(string) {
-  // your code here
-};
+  // return /[a-z]*[aeiou][a-z]*/i.test(string);
```
StuartCowley, line 103: Comments should be removed here

src/booleans.js
```diff
@@ -1,62 +1,122 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
 
 function isSquare(a) {
-  // your code here
-};
+  if (a >= 0 && Number.isInteger(Math.sqrt(a))) {
```
StuartCowley, line 87: See above about simplicity of code

src/objects.js
```diff
@@ -1,41 +1,54 @@
 const createPerson = (name, age) => {
-  // your code here
+  const obj = { name, age };
+  return obj;
```
StuartCowley, line 3: Rather than creating the variable could just return the object when its created:
```suggestion
  return { name, age };
```

src/objects.js
```diff
@@ -1,41 +1,54 @@
...
+    return object.manufacturer === 'Honda';
+  });
 };
 
 const averageAge = people => {
-  // your code here
+  const arr = Object.values(people);
+  const sum = (accumulator, obj) => accumulator + obj.age;
+  const avg = arr.reduce(sum, 0) / arr.length;
+  return avg;
```
StuartCowley, line 40: I see what you were thinking here. A few things you could do though:
- instead of creating the `sum` variable. only to pass it straight into the reduce function you could put it straight in the function itself,
- you dont need to use `Object.values` to create an array to run the reduce function on, you can just call reduce on the array being passed in (`people` is already an array if you look at the test cases),
- the creation of the `avg` variable which is just returned straight afterwards isnt needed, its value can just be returned without storing it in a variable first

Putting all this together, this is the result:
```suggestion
  return (
    people.reduce((accumulator, obj) => accumulator + obj.age, 0) /
    people.length
  );
```

src/strings.js
```diff
@@ -1,26 +1,30 @@
-function sayHello (string) {
+function sayHello(string) {
   // your code here
-};
+  return 'Hello, ' + string + '!';
```
StuartCowley, line 3: All the `// your code here` comments should be removed

StuartCowley (commented):
Good stuff, theres some interesting solutions here! Generally well done, Ive left a few comments about unnecessary variable creation, shortening the boolean functions in particular, leaving comments in place, and a few other things here and there. 
All tests pass though, so well done!

PR #6, opened 2022-03-02
Ben quinn code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,93 @@
 const getNthElement = (index, array) => {
-  // your code here
+  if (index > array.length - 1) {
+    index -= array.length;
+  }
+  return array[index];
```
StuartCowley, line 5: There is a native operator in JS which can do what you need here: remainder (`%`). Its like divide except it will return the remainder of the division instead of a decimal (EG 7 % 5 = 2)
```suggestion
  return array[index % array.length];
```

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
-  // your code here
-};
+const removeNthElement = (index, array) => array.splice(index, 1);
 
 const numbersToStrings = numbers => {
-  // your code here
+  const arr = numbers.map(item => {
+    return item.toString();
+  });
+  return arr;
```
StuartCowley, line 24: You dont need to create a variable to do this, just return the value directly:
```suggestion
  return numbers.map(item => {
    return item.toString();
  });
```

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+  const arr = strings.map(item => {
```
StuartCowley, line 28: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+    return item.toUpperCase();
+  });
+  return arr;
 };
 
 const reverseWordsInArray = strings => {
-  // your code here
+  const arr = strings.map(item => {
```
StuartCowley, line 35: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+      .split('')
+      .reverse()
+      .join('');
+  });
+  return arr;
 };
 
 const onlyEven = numbers => {
-  // your code here
+  const arr = numbers.filter(item => {
```
StuartCowley, line 45: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+    return item % 2 === 0;
+  });
+  return arr;
 };
 
 const removeNthElement2 = (index, array) => {
-  // your code here
+  const arr = array.filter((_, elementIndex) => elementIndex !== index);
```
StuartCowley, line 52: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+      items[0].includes('u') ||
+      items[0].includes('U')
+    );
+  });
+  return arr;
 };
 
 const removeSpaces = string => {
-  // your code here
+  const str = string.split(' ').join('');
```
StuartCowley, line 75: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+  return str;
 };
 
 const sumNumbers = numbers => {
-  // your code here
+  const arr = numbers.reduce((total, item) => {
```
StuartCowley, line 80: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+    return item + total;
+  }, 0);
+  return arr;
 };
 
 const sortByLastLetter = strings => {
-  // your code here
+  const arr = strings.sort((a, b) => {
```
StuartCowley, line 87: As with `numbersToStrings`, you can avoid assigning this to a variable just to return it

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+      items[0].includes('E') ||
+      items[0].includes('i') ||
+      items[0].includes('I') ||
+      items[0].includes('o') ||
+      items[0].includes('O') ||
+      items[0].includes('u') ||
+      items[0].includes('U')
+    );
+  });
+  return arr;
```
StuartCowley, line 71: This is a pretty verbose way of doing this! Filtering is the right approach, but theres an awful lot of duplication here that doesnt need to be. The cleanest way by far to do this is with a regular expression, which looks like this: /[aeiou]/i
This is a language agnostic way to do pattern recognition, which is ultimately what you are trying to do here. Basically the contents of the square brackets are the characters you want to match, and we can return only the elements we want by using the built in test function like this:
`const result = /[aeiou]/i.test(yourStringHere)`
if `yourStringHere` contains any of the values in the square brackets, it will pass. Since we need to know if just the _first_ letter is a vowel, we can add the `^` symbol before the pattern. This just means 'match this from the start of the string'.

So when combining it with filter as in the above method, you get this:
```
```
There are other ways to do this such as looping over the strings and then looping again over an array of vowels to find matches like that, but its a long way round and inefficient. Im sure you'll be seeing regex again at some point in soon, dont worry if its a bit confusing! Its one of those things that gets easier with time but is incredibly powerful once you get the hang of it

src/booleans.js
```diff
@@ -1,61 +1,62 @@
...
 };
 
 function one(a, b) {
-  // your code here
+  return !a == b;
 };
 
 function truthiness(a) {
-  // your code here
+  return !!(a);
```
StuartCowley, line 22:
```suggestion
  return !!a; 
```

src/booleans.js
```diff
@@ -1,61 +1,62 @@
...
 };
 
 function none(a, b) {
-  // your code here
+  return a == 0 && b == 0;
 };
 
 function one(a, b) {
-  // your code here
+  return !a == b;
```
StuartCowley, line 18: Very nice!
```suggestion
  return !a === b;
```

src/booleans.js
```diff
@@ -1,61 +1,62 @@
...
 };
 
 function either(a, b) {
-  // your code here
+  return a || b;
 };
 
 function none(a, b) {
-  // your code here
+  return a == 0 && b == 0;
```
StuartCowley, line 14: Should be using strict comparisons throughout this file:
```suggestion
  return a === 0 && b === 0;
```

src/booleans.js
```diff
@@ -1,61 +1,62 @@
...
 
 function startsWith(char, string) {
-  // your code here
+  return char === string[0];
 };
 
 function containsVowels(string) {
-  // your code here
+  return (string.includes("a") || string.includes("e") || string.includes("i") || 
+string.includes("o") || string.includes("u") || string.includes("A") || string.includes("E") || string.includes("I") || string.includes("O") || string.includes("U"))
```
StuartCowley, line 55: As described above, a regex is by far the cleanest solution to this. If you look closely, you should see that the pattern used here is identical to the one described in arrays (except for the missing `^`):
```suggestion
  return strings.filter(string => /[aeiou]/i.test(string));
```

src/numbers.js
```diff
@@ -1,45 +1,45 @@
...
 }
 
 function quotient (a, b) {
-  // your code here
+  return Math.trunc(a/b);
 }
 
 function remainder (a, b) {
-  // your code here
+  return a % b;
```
StuartCowley, line 42: This is the function I was talking about in `getNthElement` in arrays

StuartCowley (commented):
Good stuff! If added some info about regex which I think will prove very useful in the future, this will allow you to cut out a lot of duplication in a few of the methods here. 
Its worth mentioning that I couldnt run the tests at first, as they are all marked as skipped! Part of the instructions were to change all the `xit` blocks to `it` to allow the tests to run, I did this and they all pass so well done!

PR #7, opened 2022-03-06
Tom bogle code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,93 @@
 const getNthElement = (index, array) => {
-  // your code here
+  if (index < array.length) {
+    return array[index];
+  }
+  if (index >= array.length) {
+    return array[index - array.length];
+  }
```
StuartCowley, line 7: There is a native operator in JS which can do what you need here: remainder (%). Its like divide except it will return the remainder of the division instead of a decimal (EG 7 % 5 = 2)

```suggestion
  return array[index % array.length];
```

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+  const newArray = [];
+
+  for (let i = 0; i < strings.length; i++) {
+    newArray.push(strings[i].toUpperCase());
+  }
+  return newArray;
```
StuartCowley, line 40: While this does work, there is a cleaner way using a `map` rather than a for loop:
```suggestion
  return strings.map(string => {
    return string.toUpperCase();
  });
```
You're doing something very similar with reverseWordsInArray just below. Incidentally you don't need a higher order function here, there are built in methods in javascript as seen here that will handle this just fine

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
 };
+// Could do with refactoring to make more concise ^^ (use a higher order function) //
```
StuartCowley, line 42: Should be removed

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
-  // your code here
+  let removeNth = array.slice(0, index).concat(array.slice(index + 1, array.length));
+  return removeNth;
 };
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  let vowelStart = strings.filter(strings => /^[aeiou]/i.test(strings));
+
+  return vowelStart;
```
StuartCowley, line 72: Nice use of regex! This can be simplified a bit furhter to avoid creating a variable unnecessarily:
```suggestion
  return strings.filter(string => /^[aeiou]/i.test(string));
```

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
 
 const onlyEven = numbers => {
-  // your code here
+  let evenNumbers = [];
+  for(let i = 0; i < numbers.length; i++) {
+    if (numbers[i] % 2 === 0) {
+      evenNumbers.push(numbers[i]);
+    }
+  }
+  return evenNumbers;
```
StuartCowley, line 61: This does work, but it can be made more concise using a filter method. Most of the time arrays are manipulated using either `filter`, `map`, `forEach` or `sort` depending on what you are trying to do. In this case you are trying to remove all values that dont fit a particular criteria (those that have a remainder after dividing by 2): 
```suggestion
  return numbers.filter(number => number % 2 === 0);
```

src/arrays.js
```diff
@@ -1,61 +1,93 @@
...
+  return string.split(' ').join('');
 };
 
 const sumNumbers = numbers => {
-  // your code here
+  let sum = 0;
+  for (let i = 0; i < numbers.length; i++) {
+    sum += numbers[i];
+  }
+  return sum;
```
StuartCowley, line 84: As described above, you can use one of the most common built in array methods to reduce an array down to a single value, namely reduce:
```suggestion
  return numbers.reduce((previous, current) => previous + current, 0);
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
+  return !a;
 };
 
 function both(a, b) {
-  // your code here
+  if (a === true && b === true) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 10: If we dig a bit deeper into what this test is doing, this can be simplified a great deal. 
While the logic is sound, it could be simplified like so:
```
  if (a && b ) {
    return true;
  } else {
    return false;
  }
```
Since `a` and `b` are either true or false, you can compare them directly without having to assert whether or not they match true or false.
We can go a step further than this with the rest of the function. Since you are returning either true or false, we can just return the value of the comparison like this:
```
  return (a && b) ? true : false;
```
Furthermore, we can simplify this yet again!
```
  return a && b;
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
 };
 
 function either(a, b) {
-  // your code here
+  if (a === true || b === true) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 18: As described in the last function, this can be simplified in the same way:
```suggestion
  return a || b;
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
 };
 
 function none(a, b) {
-  // your code here
+  if (a !== true && b !== true) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 26: And again:
```suggestion
  return !(a || b);
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
+  return Boolean(a);
 };
 
 function isEqual(a, b) {
-  // your code here
+  if (a === b) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 46: Can be simplified as above:
```suggestion
  return a === b;
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
 };
 
 function one(a, b) {
-  // your code here
+  if (a === true && b === false || a === false && b === true) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 34: Can be simplified as above:
```suggestion
  return (a && !b) || (!a && b);
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
+//       const containsVowels = string => {   return Boolean(string.match(/[aeiou]/gi));   };    *regex alternative* //
+
+
+
 function isLowerCase(string) {
-  // your code here
+  if (string == string.toLowerCase()) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 116: Can be simplified as above. Also, its always best to use the strict comparison operator `===` where possible:
```suggestion
  return string === string.toLowerCase();
```

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
+  if (a === b) {
+    return true;
+  } else {
+    return false;
+  }
 };
 
 function isGreaterThan(a, b) {
-  // your code here
+  if (a > b) {
```
StuartCowley, line 50: All the functions from here on in this file could be simplified as above, would be great if these could be refactored in the same way

src/booleans.js
```diff
@@ -1,63 +1,123 @@
...
+  if (string[0] === char) {
+    return true;
+  } else {
+    return false;
+  }
 };
 
 function containsVowels(string) {
-  // your code here
+let vowels = 'aeiouAEIOU';
```
StuartCowley, line 98: Would definitely be better to use the regex alternative here, though I think what you have in the comment isnt the right way to do it. Refer to your solution for `elementsStartingWithAVowel` above (and be careful about the use of the `^` character...)

src/numbers.js
```diff
@@ -1,45 +1,50 @@
...
+    return Math.floor(a / b);
+  } 
+  if (a / b < 0) {
+    return Math.ceil(a / b);
+  }  
 }
 
 function remainder (a, b) {
-  // your code here
+  return (a % b);
```
StuartCowley, line 47: Brackets aren't needed here
```suggestion
  return a % b;
```

src/numbers.js
```diff
@@ -1,45 +1,50 @@
...
 }
 
 function quotient (a, b) {
-  // your code here
+  if (a / b > 0) {
+    return Math.floor(a / b);
+  } 
+  if (a / b < 0) {
+    return Math.ceil(a / b);
+  }
```
StuartCowley, line 43: Theres a nice built in function in JS to do just this, trunc:
```suggestion
  return Math.trunc(a / b);  
```

src/objects.js
```diff
@@ -1,41 +1,67 @@
 const createPerson = (name, age) => {
-  // your code here
+  return {
+    'name': name,
+    'age': age
```
StuartCowley, line 4: A minor point, but you can use the assignment shorthand for this:
```suggestion
    name,
    age
```

src/objects.js
```diff
@@ -1,41 +1,67 @@
...
 };
 
 const averageAge = people => {
-  // your code here
+  let avrAge = 0;
+  for (const i in people) {
+    avrAge += Math.round(people[i].age / people.length);
+  }
+  return avrAge;
+  // Could try using .reduce() as another method of getting one number from mulitple //
```
StuartCowley, line 54: Reduce is definitely the right way to go!
```
  const total = people.reduce((previous, current) => {
    return previous + current.age;
  }, 0);

  return total / people.length;
```

This could be simplified a bit further to avoid creating the variable:
```suggestion
return (
    people.reduce((previous, current) => {
      return previous + current.age;
    }, 0) / people.length
  );
```

StuartCowley (commented):
This is pretty solid, all tests pass and do their job fine. There are a few suggested improvements to making them more concise, removing unnecessary variables and using shorthand, plus making use of more appropriate built in methods to save you having to write custom functions using for loops. All the same, well done!

PR #8, opened 2023-03-06
Emma wilkinson code review

StuartCowley:


arrays.js
```diff
@@ -0,0 +1,83 @@
...
+  return string.split(",");
+};
+
+const addToArray = (element, array) => {
+  array.push(element);
+};
+
+const addToArray2 = (element, array) => {
+  let newarray = array.concat(element);
+  return newarray;
```
StuartCowley, line 19: So theres 2 things to say here, first is that you should always try and use the `const` keyword over `let` as its less prone to causing bugs caused by overwriting the value by mistake, and is the standard way of writing JS these days. The second point is that we don't actually need to create this variable in the first place, we can just return the value directly:
```
  return array.concat(element);
```

arrays.js
```diff
@@ -0,0 +1,83 @@
...
+const removeNthElement = (index, array) => {
+  return array.splice(index, 1);
+};
+
+const numbersToStrings = numbers => {
+  return numbers.map(String);
+};
+
+const uppercaseWordsInArray = strings => {
+  return strings.map(strings => strings.toUpperCase());
```
StuartCowley, line 31: Theres an eslint error happening here, which is to do with the syntax you're using for the `map` function. Hovering over the text in your IDE will show a description of the error, in this case `'strings' is already declared in the upper scope`. This is telling you are are trying to use a value for a temporary variable that already exists, which is a bad practice.
The parameter givenn to the callback function in functions like `map` or `filter` can be anything you like, but it makes sense to use something semantic that makes logical sense. In this case it could be swapped out for 'string':

```
  return strings.map(string => string.toUpperCase());
```
Problem solved

arrays.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const reverseWordsInArray = strings => {
+  return strings.map(word => word.split("").reverse().join(""));
+
+};
+
+const onlyEven = numbers => {
+  return numbers.filter(numbers => numbers % 2 === 0);
```
StuartCowley, line 40: As above, the scope issue needs rectifying:
```
```

arrays.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const removeNthElement2 = (index, array) => {
+  return array.slice(0, index).concat(array.slice(index + 1));
+};
+
+const elementsStartingWithAVowel = strings => {
+ let rex = /^[aeiou]/i;
+return results = strings.filter(strings => rex.test(strings));
```
StuartCowley, line 49: Nice to see some regex being attempted here! This works, but is a bit messy and there are some eslint errors to fix. Hovering the mouse over the error will show the problem here, you are trying to assign a value to result in the return statement, and the same scope issue detailed above is here too. You can fix this with:
```
  return strings.filter(string => rex.test(string));
```

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+  return !a;
+};
+
+function both(a, b) {
+  if (a == true && b == true) {
+    return true;
+  } else {
+    return false;
+  };
+
```
StuartCowley, line 11: So a few things to say here, eslint is giving some pretty heavy hints about what could be done a bit better with the syntax. First off is not using the strict equality operator `===` which is always the better option. Second, you can rewrite this function so it will return true based on a single condition:
```
  if (a && b) {
    return true;
  }
  return false;
```
Going a bit further you could turn it into a simple ternary operator:
```
return (a && b) ? true : false;
```
Which as it happens my editor simplified to just this!
```
  return (a && b);
```

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+function either(a, b) {
+  if (a == true | b == true) {
+
+    return true;
+  } else {
+
+    return false;
+  };
```
StuartCowley, line 21: As above, you could shorten this quite a lot by just returning the result of the initial evaluation:
```
  return a || b;
```
Look carefully at the error eslint is giving you too. What youre actually doing here is using a bitwise operator and not the 'OR' operator, which is `||` and not just `|`

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+function none(a, b) {
+  return negate(either(a, b));
```
StuartCowley, line 25: See what you did here, very efficient! If these functions were not here though, you would need a way to do this.

As it happens you can use the same logic described above:
```
  return !(a || b);
```
As a bonus this will be more performant, as it involves fewer function calls. Points for thinking round corners though and reusing existing functions, I like that

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+function one(a, b) {
+  if ((a == true | b == true) && a !== b) {
+    return true;
+  } else {
+    return false;
+  };
```
StuartCowley, line 33: As before, hits can be refactored a bit: 
```
```

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+function truthiness(a) {
+  if (a) {
+    return true;
+  } else {
+    return false;
+  };
```
StuartCowley, line 41: The logic here is fine, but it can be condensed a lot using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+function startsWith(char, string) {
+  if (string.charAt(0) == char) {
+    return true;
+  } else {
+    return false;
+  };
+};
```
StuartCowley, line 86: All these functions can be re-written in the shorter way described above

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+
+
+function isLowerCase(string) {
+return string.toLowerCase() === string;
+/*  const lowerString = string.toLowerCase(); 
+  if (string == lowerString) {
+  return true; 
+  }
+  return false;
+  Both codes do the same thing. */
```
StuartCowley, line 105: You never need to include multiple versions of the same piece of logic in your code, this comment should be removed

booleans.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+function containsVowels(string) {
+  string = string.toLowerCase();
+  const vowels = ['a', 'e', 'i', 'o', 'u'];
+  for (let i = 0; i < string.length; i++) {
+    if (vowels.includes(string[i])) return true;
+  }
+  return false;
+};
```
StuartCowley, line 95: So this feels like a good chance to expand on some use of regex you tried earlier!
This is one of those things that you'll end up using at some point, its a very versatile and powerful way of matching patterns in strings very quickly. In this case we want to use it to define our set of vowels: `/[aeiou]/`.

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise.

numbers.js
```diff
@@ -0,0 +1,61 @@
...
+
+function roundDown (a) {
+
+  return Math.floor(a);
+
+};
+
+function absolute (a) {
+  return Math.abs(a);
+};
```
StuartCowley, line 39: Why the mix of function and constant declarations here? You should try and keep this consistent across the whole project:
```
const round = a => {
  return Math.round(a);
};

const roundUp = a => {
  return Math.ceil(a);
};

const roundDown = a => {
  return Math.floor(a);
};

const absolute = a => {
  return Math.abs(a);
};
```

objects.js
```diff
@@ -0,0 +1,84 @@
+const createPerson = (name, age) => {
+  return person = {
+    name:name,
+    age:age,
+  };
```
StuartCowley, line 5: Eslint is complaining here as you can use the shorthand form for this:
```
  return {
    name,
    age
  };
```
You are also assigning a variable in the return statement. Most of the time checking the eslint message will give you the answer to issues like this, its always worth reading them

objects.js
```diff
@@ -0,0 +1,84 @@
...
+const getProperty = (property, object) => {
+  return object[property];
+};
+
+const hasProperty = (property, object) => {
+  if (object.hasOwnProperty(property) == true) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 21: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

objects.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+const isOver65 = person => {
+  if (person.age > 65) {
+    return true; 
+  } else {
+    return false;
+  }
```
StuartCowley, line 29: As above, you could simply return the result of evaluating the condition since you are only looking for a boolean value here:
```
  return person.age > 65;
```

objects.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+const getAges = people => {
+  let arrAges = []
+  for (let i = 0; i < people.length; i++) {
+      arrAges.push(people[i].age);
+    };
+  return arrAges;
+  };
```
StuartCowley, line 38: Remember your closing tags should be on the same indentation level as the opening of the function
```
}; 
```

objects.js
```diff
@@ -0,0 +1,84 @@
...
+  if (person.age > 65) {
+    return true; 
+  } else {
+    return false;
+  }
+};
+
+const getAges = people => {
+  let arrAges = []
+  for (let i = 0; i < people.length; i++) {
```
StuartCowley, line 34: Another slight point for eslint here, it doesnt like it when you use the i++ operator in for loops. Easy fixed:
```
    for (let i = 0; i < people.length; i + 1) {
```

objects.js
```diff
@@ -0,0 +1,84 @@
...
+    };
+  return arrAges;
+  }; 
+
+const findByName = (name, people) => {
+  for (let i = 0; i < people.length; i++) {
+    if (people[i].name == name) {
+      return people[i];
+    };
+  };
```
StuartCowley, line 45: You could use the built in filter method here for this:
```
```
This gets around the eslint error coming up because there is no return value at the end of the function

objects.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+const findHondas = cars => {
+  let hondaCars = []
+  for (let i = 0; i < cars.length; i++) {
+    if (cars[i].manufacturer == 'Honda') {
+      hondaCars.push(cars[i]);
+    };
+  };
+  return hondaCars;
```
StuartCowley, line 55: AS above, you could refactor this a bit using filter:
```
```

objects.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+const averageAge = people => {
+  return people.reduce((acc,a) => a.age + acc,0)/people.length;
+};
+
+const createTalkingPerson = (name, age) => {
+ const person = {
```
StuartCowley, line 63: This should be indented to 2 spaces:
```
  const person = {
```

objects.js
```diff
@@ -0,0 +1,84 @@
...
+    name,
+    age,
+    introduce: function introduction(name){
+      return 'Hi ' + name + ', my name is ' + this.name + ' and I am ' + this.age + '!';
+    }
```
StuartCowley, line 68: This is a bit messy. You can clear it up by using a template literal and also writing the function in the same style as the parent function:
```
    introduce: stranger => {
      return `Hi ${stranger}, my name is ${person.name} and I am ${person.age}!`;
    },
```

strings.js
```diff
@@ -0,0 +1,34 @@
+function sayHello (string) {
+
+  return 'Hello, ' + string + '!';
+
+};
```
StuartCowley, line 5: You dont need all these clear lines here, should just condense it down as you have with the others. Also, it would be good for these all to be written in the ES6 format (using `const` to define the functions) for a bit of consistency

StuartCowley (commented):
Theres some interesting ideas here, particularly liked how you reused some of the functions in others! Not exactly the point of the exercise but all the same wanted to say well done for doing things your own way
Ive left some suggestions around refactoring some functions that could be clearer, some things to tidy up a couple of formatting errors and fix the eslint issues.
The way you have the project set up with the git branches is very unusual, it looks like you started a branch and copied the entirety of the repo over, then created another with just your code changes. This is not really how git is supposed to be used, it would be a good idea to look into this a bit more and see how others are using it. Its designed to be used as a way to build on existing code, not recreate it with a lot of copy pasting, but this is something most people get better at as they move through the course. 
If you have any questions about it do please ask in the cohort channel :)

PR #9, opened 2023-03-08
Jenny rafferty code review

StuartCowley:


src/strings.js
```diff
@@ -1,25 +1,25 @@
...
 };
 
 function countCharacters (string) {
-  // your code here
+  return string.length;
 };
 
 function firstCharacter (string) {
-  // your code here
+return string.charAt(0);
```
StuartCowley, line 18: Should be indented to 2 spaces since youre inside a function here
```
  return string.charAt(0);
```

src/arrays.js
```diff
@@ -1,61 +1,85 @@
...
+}
 
 const csvStringToArray = string => {
-  // your code here
+  return string.split(',');
 };
 
 const addToArray = (element, array) => {
-  // your code here
+ array.push(element);
```
StuartCowley, line 14: This should be indented 2 spaces since youre inside a function:
```
  array.push(element);
```

src/arrays.js
```diff
@@ -1,61 +1,85 @@
...
-  // your code here
+  return strings.map(string =>
+    string
+      .split('')
+      .reverse()
+      .join('')
+      .split(' ')
+      .reverse()
+      .join(' ')
+  );
```
StuartCowley, line 42: You only need to do half of what you have here, this gives you the same result:
```
  return strings.map(string =>
    string
      .split('')
      .reverse()
      .join('')
  );
```

src/arrays.js
```diff
@@ -1,61 +1,85 @@
...
+      string.startsWith('e') ||
+      string.startsWith('i') ||
+      string.startsWith('o') ||
+      string.startsWith('u') ||
+      string.startsWith('A') ||
+      string.startsWith('E') ||
+      string.startsWith('I') ||
+      string.startsWith('O') ||
+      string.startsWith('U')
+  );
```
StuartCowley, line 68: The basic idea is here, but theres a few different ways to go about this. This has a lot of repetition, and isnt that maintainable so should be re-written. One way is to have an array of values to check for, then use a combination of the filter and includes array methods, and the toLowerCase string method:

src/arrays.js
```diff
@@ -1,61 +1,85 @@
...
 };
 
 const removeSpaces = string => {
-  // your code here
+  return string.replaceAll(' ', '');
```
StuartCowley, line 72: This doesnt work, as the replaceAll method needs a regex to work. Rather than going into regex here you can fix this using regular string methods:
```
```

src/booleans.js
```diff
@@ -1,61 +1,72 @@
...
+    string.includes('e') ||
+    string.includes('i') ||
+    string.includes('o') ||
+    string.includes('u') ||
+    string.includes('A') ||
+    string.includes('E') ||
+    string.includes('I') ||
+    string.includes('O') ||
+    string.includes('U')
+  );
```
StuartCowley, line 65: So this feels like a good chance to introduce regex to you!
This is one of those things that you'll end up using at some point, its a very versatile and powerful way of matching patterns in strings very quickly. They are written in a format like `/[aeiou]/`, and once you get the hang of them they are really useful

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise.

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object.hasOwnProperty(property);
```
StuartCowley, line 16: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
 
 const isOver65 = person => {
-  // your code here
+  return person.age > 65;
 };
 
 const getAges = people => {
-  // your code here
+  const ageArray = people.map(person => person.age);
+  return ageArray;
```
StuartCowley, line 25: This can be condensed to a single line:
```
```

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
 const findHondas = cars => {
-  // your code here
+  return cars.filter(car => car.manufacturer === 'Honda');
 };
 
 const averageAge = people => {
-  // your code here
+  const ageArr = people.map(person => person.age)
+  const averageAge = ageArr.reduce((a, b) => (a + b));
+  return averageAge / people.length;
```
StuartCowley, line 39: This should be rewritten a bit. Youre using the same variable name as the function internally here so eslint is complaining, even if this wasnt the case this could be shortened a bit anyhow:
```
  const ageArr = people.map(person => person.age).reduce((a, b) => a + b);
  return ageArr / people.length;
```

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
 };
 
 const createTalkingPerson = (name, age) => {
-  // your code here
+  return {
+    name: name,
+    age: age,
```
StuartCowley, line 45: You can use the object shorthand here, does exactly the same thing as you have here just makes it a bit more concise:
```
    name,
    age,
```

StuartCowley (commented):
This is pretty solid work, theres a few minor formatting problems to fix, have left some suggestions on how to improve the code in a few other places, plus have left a bit of info on regex which is something you will no doubt be coming across at some point in the future. 
For the most part great work!

PR #10, opened 2023-03-09
David tech ding code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,76 @@
 const getNthElement = (index, array) => {
-  // your code here
+  if (index >= array.length){
+    index = index % array.length;
+  }
+  return array[index];
```
StuartCowley, line 5: This works, but a small change can make this more concise using the logic on line 3:
```
```
Also reassigning a value passed in as a prop inside a function is not really a good idea as it can lead to unexpected behaviour in larger functions, as the value may not be what you expect it to be. In this case we can rewrite it to avoid the issue but its worth saying anyhow

src/arrays.js
```diff
@@ -1,61 +1,76 @@
...
 };
 
 const csvStringToArray = string => {
-  // your code here
+  return string.split(',');
 };
 
 const addToArray = (element, array) => {
-  // your code here
+   array.push(element);
```
StuartCowley, line 17: Only need 2 spaces for indentation, you have 3 here:
```
  array.push(element);
```

src/arrays.js
```diff
@@ -1,61 +1,76 @@
...
+  strings.sort((a , b) => {
+    const lastLetterA = a.slice(-1);
+        const lastletterB = b.slice(-1);
+        if (lastLetterA < lastletterB){
+          return -1;
+        } else {
+          return 1;
+        }
+      })
+      return strings;
```
StuartCowley, line 73: A couple of things here. The formatting is a bit of a mess, remember you should only go an extra layer into your indentation when going further inside a function, and it should be the same amount each time (in this case 2 spaces as with the rest of your work)
Also, eslint is complaining about the unneeded else statement. You can remove this easily enough:
```
  strings.sort((a, b) => {
    const lastLetterA = a.slice(-1);
    const lastletterB = b.slice(-1);
    if (lastLetterA < lastletterB) {
      return -1;
    }
    return 1;
  });
  return strings;
```

src/booleans.js
```diff
@@ -1,61 +1,61 @@
...
 };
 
 function isEven(a) {
-  // your code here
+  return a % 2 === 0;
 };
 
 function isSquare(a) {
-  // your code here
+  return a>= 0 && Math.sqrt(a) % 1 === 0;
```
StuartCowley, line 46: You dont need the check for greater than 0 here:
```
  return Math.sqrt(a) % 1 === 0;
```

src/booleans.js
```diff
@@ -1,61 +1,61 @@
...
 
 function both(a, b) {
-  // your code here
-};
+  return a && b;
+}
 
 function either(a, b) {
-  // your code here
+ return a || b;
```
StuartCowley, line 10: 2 spaces indentation needed here
```
  return a || b;
```

src/numbers.js
```diff
@@ -1,45 +1,49 @@
...
-  // your code here
-}
+const divide = (a, b) => {
+  return a / b;
+};
 
-function power (a, b) {
-  // your code here
+const power = (a, b) => {
+  return Math.pow(a, b);
```
StuartCowley, line 18: You have an eslint error on this line. Remember if you hover over the code with red lines below it VSCode shows some info on what the issue is and how to resolve it, in this case it tells you to use the exponentiation operator instead:
```
  return a ** b;
```

src/numbers.js
```diff
@@ -1,45 +1,49 @@
...
 }
 
-function round (a) {
-  // your code here
+const round = (a) => {
```
StuartCowley, line 21: Eslint is throwing an error here as you are using brackets the parameters when there is only one there, you can remove the brackets and and for the next few functions:
```
const round = a => {
```

src/numbers.js
```diff
@@ -1,45 +1,49 @@
...
 }
 
-function quotient (a, b) {
-  // your code here
+const quotient = (a, b) => {
+  if (a < 0 && b > 0){
+    return Math.ceil(a / b);
+  } else {
+    return Math.floor(a / b);
+  }
```
StuartCowley, line 42: Theres a built in function for this: Math.trunc (see [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#:~:text=The%20Math.,by%20removing%20any%20fractional%20digits.))
```
  return Math.trunc(a / b);
```

src/objects.js
```diff
@@ -1,41 +1,51 @@
 const createPerson = (name, age) => {
-  // your code here
+  return {
+    name: name,
+    age: age
```
StuartCowley, line 4: Eslint is complaining here as you can use the shorthand form for this:
```
    name,
    age,
```

src/objects.js
```diff
@@ -1,41 +1,51 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object.hasOwnProperty(property);
```
StuartCowley, line 17: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,51 @@
...
 
 const findHondas = cars => {
-  // your code here
+  return cars.filter(car => car.manufacturer === 'Honda');
 };
 
 const averageAge = people => {
-  // your code here
+    const totalAge = people.reduce((sum,person) => sum + person.age, 0);
+    return totalAge/people.length;
```
StuartCowley, line 38: Logic here is fine, but eslint is complaining about spacing and you should make sure you have 2 spaces of indentation at the start of each line:
```
  const totalAge = people.reduce((sum, person) => sum + person.age, 0);
```

src/objects.js
```diff
@@ -1,41 +1,51 @@
...
 };
 
 const createTalkingPerson = (name, age) => {
-  // your code here
+  return {
+    name: name,
+    age: age,
+    introduce: function(otherPersonName){
+      return `Hi ${otherPersonName}, my name is ${this.name} and I am ${this.age}!`
+    }
```
StuartCowley, line 47: Eslint has something to say here again, this time about the manner this function has been written in. An easy change though, and it even provides the solution for you when hovering over the red squiggly line in your IDE:
```
    introduce(otherPersonName) {
      return `Hi ${otherPersonName}, my name is ${this.name} and I am ${this.age}!`;
    }
```

src/strings.js
```diff
@@ -1,26 +1,26 @@
-function sayHello (string) {
-  // your code here
-};
+const sayHello = (string) => {
```
StuartCowley, line 1: Every function here has eslint errors caused by using brackets when a single parameter is passed in, they should all be removed

src/strings.js
```diff
@@ -1,26 +1,26 @@
...
-  // your code here
-};
+const countCharacters = (string) => {
+  return string.length;
+}
 
-function firstCharacter (string) {
-  // your code here
-};
+const firstCharacter = (string) =>{
```
StuartCowley, line 17: Should leave a space between the `>` and the `{` opening bracket of your function:
```
const firstCharacter = string => {
```

StuartCowley (commented):
Good work! The logic for all your functions works, though there are a few things you could look at:
- The tests should all have been updated to change all instances of `xit` to `it`, this is the default way of doing things, I had to go through and update this in order to run the tests
- Paying attention to eslint will do you a lot of favours going forward. There are quite a few errors being thrown throughout your work here, all of which are easily fixed and all related to indentation or general formatting, things like missing spaces off or indenting to the wrong level. It is there to help you and is a very useful tool, so would highly recommend getting used to using it going forward

Some good ideas and approaches to solving the functions here, keep it up :)

PR #11, opened 2023-03-10
Max tauroza code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,106 @@
 const getNthElement = (index, array) => {
-  // your code here
-};
+   index2 = index - array.length;
+  if (index >= array.length) {
+    return array[index2]
+  }
+  return array[index];
```
StuartCowley, line 6: You can simplify this quite a bit using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). 
```
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
+}
 
 const arrayToCSVString = array => {
-  // your code here
+ return array.toString();
```
StuartCowley, line 10: Should be indented to 2 spaces here :
```
  return array.toString();
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
-  // your code here
+  return array.splice(index, 1);
 };
 
 const numbersToStrings = numbers => {
-  // your code here
+  const numberString = numbers.map(number => {
+    return number.toString();
+  });
+  return numberString;
```
StuartCowley, line 33: Logic here is good, but you dont need to create the variable to just return it at the end. You could just return the result of the map function directly:
```
  return numbers.map(number => {
    return number.toString();
  });
```
If you wanted to condense it a bit more, you could use the implicit return feature of `map` and put it all on the same line:

```
  return numbers.map(number => number.toString());
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+  const uppercaseWords = strings.map(word => {
+    return word.toUpperCase();
+  });
+  return uppercaseWords;
```
StuartCowley, line 40: As above:
```
  return strings.map(word => word.toUpperCase());
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
 
 const reverseWordsInArray = strings => {
-  // your code here
+  const stringsToArrays = strings.map(element => {
+    return element.split('').reverse();
+  });
+  const backToStrings = stringsToArrays.map(arrayElement => {
+    return arrayElement.toString().replaceAll(',', '');
+  });
+  return backToStrings;
```
StuartCowley, line 50: This doesnt work, though I see what you were trying to do. It falls down on line 47 where you are trying to use replaceAll on an array instead of a string. Instead, think about what you want to do again:
- for each of the strings, you want to split it into an array
- reverse these,
- stick them back together now reversed.

if it was a single string being passed in to the function, you could do this like so:
```
string
  .split('')
  .reverse()
  .join('')
```

But since what we are passing in is an array of strings, we want to do this same process for each of them, so we can wrap all that inside a `map` function:

```
  return strings.map(string =>
    string
      .split('')
      .reverse()
      .join('')
  );
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
 };
 
 const onlyEven = numbers => {
-  // your code here
+  const evenNumbers = numbers.filter(number => {
+    return number % 2 === 0;
+  });
+  return evenNumbers;
```
StuartCowley, line 57: As above this can be shortened to a one liner:
```
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
 };
 
 const removeNthElement2 = (index, array) => {
-  // your code here
+  const firstHalf = array.slice(0, index);
+  const secondHalf = array.slice(index + 1);
+  const newArray = firstHalf.concat(secondHalf);
+  return newArray;
```
StuartCowley, line 64: This could be shortened a bit without sacrificing readability:
```
  const firstHalf = array.slice(0, index);
  const secondHalf = array.slice(index + 1);
  return firstHalf.concat(secondHalf);
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  const firstLetter = strings.filter(string =>
+      string.toLowerCase().startsWith('a') ||
+      string.toLowerCase().startsWith('e') ||
+      string.toLowerCase().startsWith('i') ||
+      string.toLowerCase().startsWith('o') ||
+      string.toLowerCase().startsWith('u')
+  );
+  return firstLetter;
```
StuartCowley, line 75: The basic idea is here, but theres a few different ways to go about this. This has a lot of repetition, and isnt that maintainable so should be re-written. One way is to have an array of values to check for, then use a combination of the filter and includes array methods, and the toLowerCase string method:

```

```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
 };
 
 const removeSpaces = string => {
-  // your code here
+  return string.replaceAll(' ', '');
```
StuartCowley, line 79: This doesnt work, as the replaceAll method needs a regex to work. Rather than going into regex here you can fix this using regular string methods:
```
```

src/arrays.js
```diff
@@ -1,61 +1,106 @@
...
+  });
+  const arraysSorted = stringsToArrays2.sort();
+  const finalArray = arraysSorted.map(sortedArr => {
+    return sortedArr
+      .split('')
+      .reverse()
+      .toString()
+      .replaceAll(',', '');
+  });
+  return finalArray;
```
StuartCowley, line 103: This is an interesting approach. Unfortunately it doesnt work as you cant use the `replaceAll` function here. There are multiple ways to do this, one would be to use the sort function:
```
  return strings.sort((a, b) => a.charCodeAt(a.length - 1) - b.charCodeAt(b.length - 1));
```

src/booleans.js
```diff
@@ -1,62 +1,77 @@
...
 
 function truthiness(a) {
-  // your code here
-};
+  if (a) {
+    return true;
+  }
+  if (!a) {
+    return false;
+  }
```
StuartCowley, line 27: A couple of things here, first is that eslint is unhappy because you dont have a return value at the end of the function. You could fix this by removing the unnecessary second `if` statement:
```
  if (a) {
    return true;
  }
  return false;
```
This is still unnecessarily long though. A nice trick is the use of two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/booleans.js
```diff
@@ -1,62 +1,77 @@
...
+  return a <= b;
+}
 
 function isOdd(a) {
-  // your code here
-};
+  if (a % 2 !== 0) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 46: You can update a lot of these functions to be more concise by simply returning the evaluation of the condition, for example here:
```
  return a % 2 !== 0;
```
If it evaluates to `true`, you get `true` returned and vice versa

src/booleans.js
```diff
@@ -1,62 +1,77 @@
...
+}
 
 function isEven(a) {
-  // your code here
-};
+  if (a % 2 === 0) {
```
StuartCowley, line 50: You should update this as above

src/booleans.js
```diff
@@ -1,62 +1,77 @@
...
 function isSquare(a) {
-  // your code here
-};
+  return Number.isInteger(Math.sqrt(a));
+}
 
 function startsWith(char, string) {
-  // your code here
-};
+  if (char === string.charAt(0)) {
```
StuartCowley, line 61: Again, quick update as above to make it into a clean one liner

src/booleans.js
```diff
@@ -1,62 +1,77 @@
...
-  // your code here
-};
+  return (/[aeiou]/gi.test(string))
+}
 
 function isLowerCase(string) {
-  // your code here
-};
+  if (string == string.toLowerCase()) {return true}
+  else {return false}
```
StuartCowley, line 73: Eslint did not like this. Its good practice to always make sure your code is formatted properly as it just make things easier for everyone all round (yourself included!). It could also be re-written a bit:
```
```
Note the triple equals in the middle here, where possible you should always use strict equality for comparisons

src/numbers.js
```diff
@@ -1,45 +1,45 @@
 function add (a, b) {
-  // your code here
+  return (a+b)
```
StuartCowley, line 2: The logic in all of these is sound, but the formatting is off in all of them. You should have a clear space between each argument in the return statement and no brackets, so here:
```
  return a + b;
```

src/objects.js
```diff
@@ -1,41 +1,62 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object.hasOwnProperty(property);
```
StuartCowley, line 18: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,62 @@
...
-  // your code here
+  return people.map(personAge => {
+    return personAge.age;
+  });
 };
 
 const findByName = (name, people) => {
-  // your code here
+  const result = people.find(person => person.name === name);
+  return result;
```
StuartCowley, line 33: No need to create a variable here:
```
  return people.find(person => person.name === name);
```

src/objects.js
```diff
@@ -1,41 +1,62 @@
...
 };
 
 const findHondas = cars => {
-  // your code here
+  const result = cars.filter(car => car.manufacturer === 'Honda');
+  return result;
```
StuartCowley, line 38: As above
```
```

src/objects.js
```diff
@@ -1,41 +1,62 @@
...
 
 const averageAge = people => {
-  // your code here
+  const ageArray = people.map(personAge => {
+    return personAge.age;
+  });
+  const totalAge = ageArray.reduce((acc, val) => {
+    return acc + val;
+  });
+  return totalAge / people.length;
```
StuartCowley, line 48: This works, though you can refactor it a bit to make if a little cleaner and chain the various methods together:
```
  const totalAge = people
    .map(personAge => personAge.age)
    .reduce((acc, val) => {
      return acc + val;
    });

```

src/strings.js
```diff
@@ -1,25 +1,25 @@
...
 };
 
 function firstCharacter (string) {
-  // your code here
+  return string.charAt(0)
 };
 
 function firstCharacters (string, n) {
-  // your code here
+  return string.slice(0, (0+n))
```
StuartCowley, line 22: Spacing is a bit off here:
```
  return string.slice(0, 0 + n);
```

StuartCowley (commented):
Theres a few tings failing here, is see what you were trying to do in each case though and its good you at least made an attempt at them! Have added explanations for why these didnt work and how to fix them
Have also left some comments about formatting, cleaning up some of the working functions, indentation and spacing, and a couple of other bits here and there too. Generally good work, looks like a solid start for the most part!

PR #12, opened 2023-03-14
Daniel mossop code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,68 @@
...
+  return strings.map(string => string.toUpperCase());
 };
 
 const reverseWordsInArray = strings => {
-  // your code here
+  return strings.map(string => string
+      .split('')
+      .reverse()
+      .join('')
+  );
```
StuartCowley, line 37: Slight formatting tweak here:
```
  return strings.map(string =>
    string
      .split('')
      .reverse()
      .join('')
  );
```

src/arrays.js
```diff
@@ -1,61 +1,68 @@
...
-  // your code here
+  const arrayTwo = array.slice();
+  arrayTwo.splice(index, 1)
+  return arrayTwo;
 };
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  const vowels = ['a', 'e', 'i', 'o', 'u', 'A', 'E', 'I', 'O', 'U'];
+  return strings.filter(string => vowels.includes(string[0]));
```
StuartCowley, line 52: The basic idea is here, but theres a few different ways to go about this. This has more repetition than is necessary, we could use a combination of the filter and includes array methods as you are doing but also include the `toLowerCase` string method:

```

```
A bit more concise and does the same job

src/arrays.js
```diff
@@ -1,61 +1,68 @@
...
 };
 
 const removeSpaces = string => {
-  // your code here
+  return string.replaceAll(' ', '');
```
StuartCowley, line 56: This doesnt work, as the replaceAll method needs a regex to work. Rather than going into regex here you can fix this using regular string methods:
```
```

src/arrays.js
```diff
@@ -1,61 +1,68 @@
...
 };
 
 const sumNumbers = numbers => {
-  // your code here
-};
-
+  return numbers.reduce((a, b) => a + b);
+}
```
StuartCowley, line 61: There should always be a clear line between each of your function definitions

src/booleans.js
```diff
@@ -1,61 +1,77 @@
...
 };
 
 function isEqual(a, b) {
-  // your code here
+  return a === b;
 };
 
 function isGreaterThan(a, b) {
-  // your code here
+  return a>b;
```
StuartCowley, line 30: There should be a space between your variables and operands in your functions, a few more instances of this below
```
  return a > b;
```

src/booleans.js
```diff
@@ -1,61 +1,77 @@
...
 };
 
 function isLessThanOrEqualTo(a, b) {
-  // your code here
+  return a<=b;
```
StuartCowley, line 34:
```
  return a <= b;
```

src/booleans.js
```diff
@@ -1,61 +1,77 @@
...
+    string.includes ('E') ||
+    string.includes ('I') ||
+    string.includes ('O') ||
+    string.includes ('U') 
+  );
+    
+    
+
+
+
```
StuartCowley, line 70: So this feels like a good chance to introduce regex to you!
This is one of those things that you'll end up using at some point, its a very versatile and powerful way of matching patterns in strings very quickly. They are written in a format like `/[aeiou]/`, and once you get the hang of them they are really useful

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise

Also you never need more than one clear line in your code, all these extra blank lines at the bottom dont serve any purpose

src/objects.js
```diff
@@ -1,41 +1,50 @@
 const createPerson = (name, age) => {
-  // your code here
+  this.name = name;
+  this.age = age;
+  return { name, age };
```
StuartCowley, line 4: Using the `this` keyword here doesnt make any sense, as you arent using the variables you just assigned in a JS object style way in the return statement. You could re-write this just to:
```
  return {
    name,
    age
  };
```

src/objects.js
```diff
@@ -1,41 +1,50 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object.hasOwnProperty(property);
```
StuartCowley, line 16: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,50 @@
...
+  const myArr = people.map(person => person.age).reduce((a, b) => a + b);
+  return myArr / people.length;
 };
 
 const createTalkingPerson = (name, age) => {
-  // your code here
+  return {
+    name,
+    age,
+    introduce(friend){
```
StuartCowley, line 44: There should be a space between the final ) and the opening { here"
```
    introduce(friend) {
```

StuartCowley (commented):
Good stuff! Everything passed the tests except for the removeSpaces method. Have left some suggestions about how to fix this as well as improve the readability and formatting, some alternate ways to do the same things you here, a couple of cleanups and an introduction to regex for string methods

Generally though solid work, keep it up :)

src/arrays.js
possomd: Hi Stu thanks for that, just wondering what makes it pass the Uppercase Vowels as well?

StuartCowley: Its all to do with the `string[0].toLowerCase()` bit on the second line. Thats taking the first character of the string being looped over, and converting it to lower case. This means it doesnt matter if it was upper or lower case to start with, as it will always end up in lower case, meaning the array of vowels we're using to search through only needs the lower case versions.
Does that make sense?

PR #13, opened 2023-04-28
Billy walker code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,66 @@
 const getNthElement = (index, array) => {
-  // your code here
+  return array[index];
```
StuartCowley, line 2: You have a failing test here (currently marked as skipped), where if n is greater than the number of elements it cycles back to the start.

You can fix this using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). 
```
```
So of the array had 5 items in it, and the given index was 7, this would become `array[2]` and return the right answer

src/arrays.js
```diff
@@ -1,61 +1,66 @@
...
 
 const removeNthElement2 = (index, array) => {
-  // your code here
+  const newArray = array.filter((element, idx) => idx !== index);
+  return newArray;
 };
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  return strings.filter(words => /^[aeiou]/i.test(words));
```
StuartCowley, line 47: Nice to see regex here!

src/arrays.js
```diff
@@ -1,61 +1,66 @@
...
+  return string.split(' ').join('');
 };
 
 const sumNumbers = numbers => {
-  // your code here
+  let sum = 0;
+  for (let i = 0; i < numbers.length; i++) {
+    sum += numbers[i];
+  }
+  return sum;
```
StuartCowley, line 59: While this works, there is a built in function that will do what you want here right out of the box.
This is a perfect use case for the [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) function:
```
  return numbers.reduce((accumulated, current) => accumulated + current);
```
The docs are pretty clear on this one, it does come up from time to time so I would recommend getting familiar with it!

src/booleans.js
```diff
@@ -1,61 +1,67 @@
...
-  // your code here
+  return Math.sqrt(a) % 1 === 0;
 };
 
 function startsWith(char, string) {
-  // your code here
+  if (char === string[0]) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 53: This can be simplified a bit. Since the expression `char === string[0]` will return a boolean anyway, theres no need for the extra logic here and you can just return the result of the expression:
```
  return char === string[0];
```

src/booleans.js
```diff
@@ -1,61 +1,67 @@
...
 };
 
 function containsVowels(string) {
-  // your code here
+  if (string.match(/[a|e|i|o|u|]/i, 0)) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 60: As above, you can simplify this a bit. The `.match` function in the way youre using it here will either return the first instance it finds, or `null` if it finds no matches, so we can turn this into a boolean using the double negation trick (!!) that you are using for the truthiness function above:
```
  return !!string.match(/[a|e|i|o|u|]/i, 0);
```
Here if the result is a match, we dont want to return the value but a `true` boolean, so we can use a single ! to turn that truthy value false, then another to flip it back to true as we want.
If there is no match it will return `null` (a falsy value but _not strictly equal to false_), by the same logic we can flip it to truthy and back to falsy and give us the result we want :)

src/objects.js
```diff
@@ -1,41 +1,78 @@
 const createPerson = (name, age) => {
-  // your code here
+  const newPerson = new Object();
+  newPerson.name = name;
+  newPerson.age = age;
+
+  return newPerson;
```
StuartCowley, line 6: There is a cleaner way to do this, you could just return a new object without declaring it in a variable:
```
  return {
    name: name,
    age: age
  };
```
This can be shortened using the modern syntax (since the key and value are the same here) to this:
```
  return {
    name,
    age
  };
```

src/objects.js
```diff
@@ -1,41 +1,78 @@
...
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  if (object.hasOwnProperty(property)) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 21: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,78 @@
...
 };
 
 const isOver65 = person => {
-  // your code here
+  if (person.age > 65) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 28: As above, you could simply return the result of evaluating the condition since you are only looking for a boolean value here:
```
  return person.age > 65;
```

src/objects.js
```diff
@@ -1,41 +1,78 @@
...
 
 const findByName = (name, people) => {
-  // your code here
+  for (let i = 0; i < people.length; i++) {
+    const person = people[i];
+    if (person.name === name) {
+      return person;
+    }
+  }
+  return null;
```
StuartCowley, line 42: You could use the built in filter method here for this:
```
```
This gets around the eslint error coming up because there is no return value at the end of the function

src/objects.js
```diff
@@ -1,41 +1,78 @@
...
-  // your code here
+  const Hondas = [];
+
+  for (let i = 0; i < cars.length; i++) {
+    if (cars[i].manufacturer === 'Honda') {
+      Hondas.push(cars[i]);
+    }
+  }
+
+  return Hondas;
```
StuartCowley, line 54: As above, the filter method can do this for us:
```
```

src/objects.js
```diff
@@ -1,41 +1,78 @@
...
 };
 
 const averageAge = people => {
-  // your code here
+  const average =
+    people.reduce((total, next) => total + next.age, 0) / people.length;
+  return average;
```
StuartCowley, line 60: You can just return the result of the function here, no need for the extra variable to be created:
```
  return people.reduce((total, next) => total + next.age, 0) / people.length;
```

src/objects.js
```diff
@@ -1,41 +1,78 @@
...
+  }
+
+  return {
+    name,
+    age,
+    introduce
+  };
+
+  // okay, so I need to set it so that the name and the age in the text are filled dynamically?
+  // and also the name and age of the people, too?
```
StuartCowley, line 75: You can remove the comment here! Also you can condense this a little into this:
```
  return {
    name,
    age,
    introduce: otherPersonName => {
      return `Hi ${otherPersonName}, my name is ${name} and I am ${age}!`;
    }
  };
```

StuartCowley (commented):
Good work! Theres a few things that could be refactored to make a bit cleaner, and one of the tests wasnt passing when I unskipped it but have left a fix for you to see how it can be done. This is nicely laid out, indentation and formatting is good, and the logic does the job its supposed to. Keep this up, really good to see things like regex and sort functions being used!

PR #14, opened 2023-05-02
Christina soare code review

StuartCowley:


src/arrays.js
```diff
@@ -1,63 +1,127 @@
 const getNthElement = (index, array) => {
-  // your code here
-};
+  return array [index >= array.length ? index % array.length : index]
```
StuartCowley, line 2: This is a bit more verbose than it needs to be. You have the basic logic there, but it can be shortened to just this:
```
You can simplify this quite a bit using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). 
```
```
```

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
+}
 
 const arrayToCSVString = array => {
-  // your code here
+  return array.join();
+  
+  //creates a new string from an array by concatanating all the elements of the array; by default it 
+  //uses a comma as a separator, but you can specify what separators you want to be added between the items. eg join() - comma but no space, 
+  //join(, ), join (-), join ('') no space, no comma , join (' ') space but no comma
```
StuartCowley, line 10: These seem like notes to yourself more than anything. Its good that youre taking these, but I would put them somewhere else such as a notion or google doc rather than a repo. These are for code, and filling them out like this doesnt look good to future employers and also, if you continuing taking notes in this fashion you would have all your notes and hints to yourself scattered all over the place, so it just makes more sense for them to be in one location going forward.

Notion is actually excellent for note taking, would recommend looking into it

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
 
+// .splice () changes an array by removing or replacing existing elements 
+// parameters - (start, deleteCount, item1,...,item n)
+// start - index at which to start changing the array
+// deleteCount - integer indicating the number of elements to remove from start
+// the elements to add to the array beginning from start
+
 const numbersToStrings = numbers => {
-  // your code here
+ return numbers.map(number =>number.toString())
```
StuartCowley, line 52: This should be indented to 2 spaces:
```
  return numbers.map(number => number.toString())
```

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
 };
 
+// map creates a new array populated with the results of calling a function to every element
+//in the calling array
+
 const uppercaseWordsInArray = strings => {
-  // your code here
-};
+ return strings.map(string => string.toUpperCase())
```
StuartCowley, line 59: 2 spaces of indentation wanted here:
```
  return strings.map(string => string.toUpperCase())
```

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
 };
 
+ //returns an array with the nth element removed, and does not mutate the original 
+ //.slice() makes a copy of the original array
+ //. .splice(index,1) removes 1 element at the specified index x
+
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  return strings.filter(string => string.toLowerCase().startsWith('a') || string.toLowerCase().startsWith('e') || string.toLowerCase().startsWith('i') || string.toLowerCase().startsWith('o') || string.toLowerCase().startsWith('u'))
+
```
StuartCowley, line 91: The basic idea is here, but theres a few different ways to go about this. This has a lot of repetition, and isnt that maintainable so should be re-written. One way is to have an array of values to check for, then use a combination of the filter and includes array methods, and the toLowerCase string method:

```

```

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  return strings.filter(string => string.toLowerCase().startsWith('a') || string.toLowerCase().startsWith('e') || string.toLowerCase().startsWith('i') || string.toLowerCase().startsWith('o') || string.toLowerCase().startsWith('u'))
+  
 };
 
 const removeSpaces = string => {
-  // your code here
+  return string.split(' ').join('');
+
```
StuartCowley, line 96: Dont need this extra clear line here

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
 };
 
+//split(' ') creates an array with the string words as elements; inverted commas and space (' ') instruct split() to separate string into elements 
+//where there is a space in the string
+//join ('') joins the elements of the array into a string, inverted commas with no space instructs join() to merge elements with no space between, 
+
 const sumNumbers = numbers => {
-  // your code here
+  return numbers.reduce((accumulator, number) => accumulator + number)
+
```
StuartCowley, line 105: Clear line not needed here

src/arrays.js
```diff
@@ -1,63 +1,127 @@
...
 };
 
+// .reduce method executes a callback fn on every element of the array and returns one value 
+
 const sortByLastLetter = strings => {
-  // your code here
+  function compareFunction(a, b) {
+    return a.charCodeAt(a.length-1) - b.charCodeAt(b.length-1)
+  }
+  return strings.sort(compareFunction)
```
StuartCowley, line 114: This could be written functionally like this:
```
  return strings.sort((a, b) => {
    return a.charCodeAt(a.length - 1) - b.charCodeAt(b.length - 1);
  });
```

src/booleans.js
```diff
@@ -1,62 +1,107 @@
...
+  return (a || b) && !(a && b);
+}
 
 function truthiness(a) {
-  // your code here
+  if (a) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 27: The logic here is fine, but it can be condensed a lot using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/booleans.js
```diff
@@ -1,62 +1,107 @@
...
 };
 
 function isEqual(a, b) {
-  // your code here
+  if (a===b){
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 35: You can update a lot of these functions to be more concise by simply returning the evaluation of the condition, for example here:
```
return (a === b)
```
If it evaluates to `true`, you get `true` returned and vice versa

src/booleans.js
```diff
@@ -1,62 +1,107 @@
...
 
 function startsWith(char, string) {
-  // your code here
+  const a = char.length;
+  const b = string.substring(0,a);
+  if (char===b) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 85: This works, but another approach is use the array syntax to see if the first character matches the `char` value like this:
```
This can be simplified a bit. Since the expression `char === string[0]` will return a boolean anyway, theres no need for the extra logic here and you can just return the result of the expression:
```
  return char === string[0];
```
```

src/booleans.js
```diff
@@ -1,62 +1,107 @@
...
 };
 
 function containsVowels(string) {
-  // your code here
+  const n = string.search(/[aeiou]/i);
+  if (n >= 0) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 94: You can restructure this a bit using the exact same logic youre already using to remove the unnecessary variable being created and the if statement:
```
  return string.search(/[aeiou]/i) >= 0;
```

src/booleans.js
```diff
@@ -1,62 +1,107 @@
...
 
 function isLowerCase(string) {
-  // your code here
-};
+  const strToLowerCase = string.toLowerCase();
+  if (strToLowerCase === string) {
+    return true;
+  } else {
+    return false;
+  }
```
StuartCowley, line 103: As above, this can be simplified in format a bit:
```
  return string.toLowerCase() === string;
```

src/numbers.js
```diff
@@ -1,45 +1,48 @@
...
 
 function divide (a, b) {
-  // your code here
+  return a / b;
 }
 
 function power (a, b) {
-  // your code here
+  return Math.pow(a,b);
+
```
StuartCowley, line 19: You have an eslint error on this line. Remember if you hover over the code with red lines below it VSCode shows some info on what the issue is and how to resolve it, in this case it tells you to use the exponentiation operator instead:
```
  return a ** b;
```
Also no need for the extra blank line underneath

src/numbers.js
```diff
@@ -1,45 +1,48 @@
 function add (a, b) {
-  // your code here
+    return a+b;
```
StuartCowley, line 2: You should run your code through a formatter before submitting, there should be 2 spaces of indentation here and for all the contents of the functions throughout the rest of this file

src/numbers.js
```diff
@@ -1,45 +1,48 @@
...
+  return Math.abs(a);
+  
 }
 
 function quotient (a, b) {
-  // your code here
-}
+  const c = a/b;
+  return Math.trunc(c);
+  }
```
StuartCowley, line 42: This should be put into one line, you dont need the extra variable here:
```
  return Math.trunc(a / b);
}
```

src/objects.js
```diff
@@ -1,43 +1,59 @@
 const createPerson = (name, age) => {
-  // your code here
-};
+  return {
+    'name': name,
+    'age': age
```
StuartCowley, line 4: Eslint is complaining here as you can use the shorthand form for this:
```
    name,
    age,
```
Remember that in objects, the keys (left of the `:` character) should not be wrapped in quotes, just written as plain text. The above is the shorthand form of 
```
  name: name,
  age: age
```

src/objects.js
```diff
@@ -1,43 +1,59 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object.hasOwnProperty(property);
```
StuartCowley, line 17: You can fix the eslint error being thrown here by using the array style notation for accessing the properties of the object as follows:
```
  return object[property];
```

```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,43 +1,59 @@
...
 };
 
 const isOver65 = person => {
-  // your code here
+ return person.age > 65 ? true : false;
```
StuartCowley, line 21: As above, you could simply return the result of evaluating the condition since you are only looking for a boolean value here:
```
  return person.age > 65;
```

src/objects.js
```diff
@@ -1,43 +1,59 @@
...
+  return cars.filter(car => car.manufacturer === 'Honda');
 };
 
 const averageAge = people => {
-  // your code here
+  const agesArray = people.map(person => person.age);
+  const agesTotal = agesArray.reduce((accumulator, age) => accumulator + age);
+  return agesTotal / people.length;
+
+
```
StuartCowley, line 41: Dont need these 23 extra clear lines here

src/objects.js
```diff
@@ -1,43 +1,59 @@
...
-  // your code here
+const createTalkingPerson = (name, age) => { 
+  return {
+    'name' : name,
+    'age' : age,
+    'introduce' : otherName => {
+      return `Hi ${otherName}, my name is ${name} and I am ${age}!`;
+    }
+    }
+
```
StuartCowley, line 52: Just indentation and the same issue with the keys for objects as mentioned above here
```
  return {
    name,
    age,
    'introduce' : otherName => {
      return `Hi ${otherName}, my name is ${name} and I am ${age}!`;
    }
  }
```

src/strings.js
```diff
@@ -1,25 +1,29 @@
...
 
 function lowercase (string) {
-  // your code here
+  return string.toLowerCase();
 };
 
 function countCharacters (string) {
-  // your code here
+  const str = string;
+  return str.length;
```
StuartCowley, line 15: Oneliner time!
```
  return string.length; 
```

src/strings.js
```diff
@@ -1,25 +1,29 @@
...
 function countCharacters (string) {
-  // your code here
+  const str = string;
+  return str.length; 
 };
 
 function firstCharacter (string) {
-  // your code here
+  const str = string;
+  return str.charAt(0);
```
StuartCowley, line 20: As above:
```
  return string.charAt(0);
```

src/strings.js
```diff
@@ -1,25 +1,29 @@
...
 };
 
 function firstCharacters (string, n) {
-  // your code here
+  const str = string;
+  return str.substring(0,n);
+
```
StuartCowley, line 26: As above:
```
  return string.substring(0,n);
```

StuartCowley (commented):
Good work here, all tests pass and your code is pretty clean and easy to read. Theres a few places where it could be improved by getting rid of unneeded variables, if statements and a few other things, but its mostly cosmetic. Speaking of it would be great if you could make sure to check the formatting of your code before submission, theres a few places here that needed a bit of tweaking where it shouldbt have

All in all though good work! Please keep submitting, happy to do these for you

src/arrays.js
acristinasoare: Oh okay, never thought about the prospect of a future employer having a look at these. Is it okay to edit the files after receiving feedback from you?

PR #15, opened 2023-05-02
Lois a code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,88 @@
 const getNthElement = (index, array) => {
   // your code here
+  const newIndex = index % array.length;
+  return array[newIndex];
```
StuartCowley, line 4: You could condense this down to a single line:
```
```

src/arrays.js
```diff
@@ -1,61 +1,88 @@
...
 };
 
 const removeNthElement = (index, array) => {
   // your code here
+  array.splice(index, 1);
 };
 
 const numbersToStrings = numbers => {
   // your code here
+  return numbers.map(x => String(x));
```
StuartCowley, line 36: Its a good habit to get into to use more descriptive variable names rather rather something super generic like `x`:
```
  return numbers.map(number => String(number));
```

src/arrays.js
```diff
@@ -1,61 +1,88 @@
...
 };
 
 const uppercaseWordsInArray = strings => {
   // your code here
+  return strings.map(x => x.toUpperCase());
```
StuartCowley, line 41: As above, should use a clearer variable name here

Rather than saying the same thing everywhere else in the repo, anywhere youre using something abstract like `x` for a variable like this should be updated

src/booleans.js
```diff
@@ -1,61 +1,77 @@
...
 
 function isLessThanOrEqualTo(a, b) {
   // your code here
+  return a <= b;
 };
 
 function isOdd(a) {
   // your code here
+  return a % 2 == 1;
+
```
StuartCowley, line 49: Dont need this extra blank line here

src/booleans.js
```diff
@@ -1,61 +1,77 @@
...
 };
 
 function containsVowels(string) {
   // your code here
+  return string.match(/[aeiou]/gi) !== null;
 };
 
 function isLowerCase(string) {
   // your code here
+  return string === String(string).toLowerCase();
```
StuartCowley, line 74: This can be tweaked a bit, you dont need to cast it to a string type and just do this:
```
```

src/numbers.js
```diff
@@ -1,45 +1,56 @@
...
 }
 
 function absolute (a) {
   // your code here
+  return Math.abs(a);
 }
 
 function quotient (a, b) {
   // your code here
+  return (a-(a%b))/b;
```
StuartCowley, line 47: Theres a built in function for this: Math.trunc (see [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#:~:text=The%20Math.,by%20removing%20any%20fractional%20digits.))
```
  return Math.trunc(a / b);
```

src/objects.js
```diff
@@ -1,41 +1,67 @@
...
 };
 
 const isOver65 = person => {
   // your code here
+  return person.age > 65;
 };
 
 const getAges = people => {
   // your code here
+  return people.map(x => x.age);
```
StuartCowley, line 35: As before, it would be much better to use a better variablle name in place of `x`

StuartCowley (commented):
This is really good work! 
Nice usage of a lot of methods, including some that are only available in later versions of Node. Have left some minor refactoring type suggestions throughout
A big point to note is you should definitely remove all the `// your code here` placeholder comments from the entire repo, theyre meant to be replaced with your code :) 

Its also worth mentioning that just copying the track material for the readme isnt much help. At the least it should contain how to install the project, a brief summary of what its for, how to use and test it, and your name.

Generally this is really good stuff here, keep it up!

PR #16, opened 2023-05-05
Umar farooq code review

StuartCowley:


.package.json.swo
StuartCowley, line 1: I dont know what this file is doing here, it should be removed

.package.json.swp
StuartCowley, line 1: As above, this file should not be in the repo

src/arrays.js
```diff
@@ -1,61 +1,83 @@
 const getNthElement = (index, array) => {
-  // your code here
+  index %= array.length;
+  return array[index];
```
StuartCowley, line 3: Eslint doesnt like this because you are reassigning the index parameter inside the function, you should never do this as it can lead to unexpected behaviour in larger functions. You can use basically the exact same logic you have here in a single line and not do this reassignment:
```
```

src/arrays.js
```diff
@@ -1,61 +1,83 @@
...
-  // your code here
+  return numbers.map(String);
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+  return strings
+  .join()
+  .toUpperCase()
+  .split(',');
```
StuartCowley, line 35: Just the indentation is out here:
```
  return strings
    .join()
    .toUpperCase()
    .split(',');
```

src/arrays.js
```diff
@@ -1,61 +1,83 @@
...
-  // your code here
+  return string.split(',');
 };
 
-const addToArray = (element, array) => {
-  // your code here
-};
+function addToArray(element, array) {
+  array.push(element);
+}
```
StuartCowley, line 16: Why change the format of the function? Should be the same as the others so:
```
  array.push(element);
};
```

src/booleans.js
```diff
@@ -1,61 +1,63 @@
...
 function isEven(a) {
-  // your code here
+  return !(Boolean(a % 2));
 };
 
 function isSquare(a) {
-  // your code here
+  const squareRoot = Math.sqrt(a);
+  const isPerfectSquare = Math.ceil(squareRoot) === Math.floor(squareRoot);
+  return isPerfectSquare;
```
StuartCowley, line 48: This could be condensed down to get rid of the unnecessary variable creation:
```
  return Math.ceil(squareRoot) === Math.floor(squareRoot);
```

src/booleans.js
```diff
@@ -1,61 +1,63 @@
...
 };
 
 function startsWith(char, string) {
-  // your code here
+  return string.startsWith(char);
```
StuartCowley, line 52: This is a really unusual approach!
Calling a function recursively like this is not something you see or want to do very often, in this case it works as the `string` value is still in scope when calling the internal function and its a very simple operation, in a larger case this would be much more likely to fail. One reason for this is there is no default value for the second parameter, also what if this were required for the function to work?
An alternative would be simply
```
  return char === string[0];
```

src/objects.js
```diff
@@ -1,41 +1,63 @@
...
+  return person.age > 65;
 };
 
 const getAges = people => {
-  // your code here
+  const ages = [];
+  for (let i = 0; i < people.length; i++) {
+    ages.push(people[i].age);
+  }
+  return ages;
```
StuartCowley, line 30: This could be re-written to make use of built in methods like filter and map to the same job for you  a bit more concisely:
```
  return people.filter(person => person.name).map(person => person.age);
```

src/objects.js
```diff
@@ -1,41 +1,63 @@
...
 };
 
 const findByName = (name, people) => {
-  // your code here
+  const foundName = people.find(person => person.name === name);
+  return foundName;
```
StuartCowley, line 35: This could be shortened to one line:
```
  return people.find(person => person.name === name);
```

src/objects.js
```diff
@@ -1,41 +1,63 @@
...
 };
 
 const findHondas = cars => {
-  // your code here
+  const findHonda = cars.filter(car => car.manufacturer === 'Honda');
+  return findHonda;
```
StuartCowley, line 40: As above:
```
```

src/objects.js
```diff
@@ -1,41 +1,63 @@
...
 };
 
 const averageAge = people => {
-  // your code here
+  const ages = [];
+  for (let i = 0; i < people.length; i++) {
+    ages.push(people[i].age);
+  }
+  const totalAge = ages.reduce((acc, curr) => acc + curr, 0);
+  return totalAge / people.length;
```
StuartCowley, line 49: This works, though you can refactor it a bit to make if a little cleaner and chain the various methods together:
```
  const totalAge = people
    .map(personAge => personAge.age)
    .reduce((acc, val) => {
      return acc + val;
    });

```

StuartCowley (commented):
Great work! Theres some solid logic here, have left a few suggestions to improve things in places but nothing serious

There are a couple of things which are a bit unusual that should be mentioned, you have 2 extra files which dont have any purpose, and your readme should be updated to detail what your repo is for, how its installed and tested, and what functionality it covers

PR #17, opened 2023-05-09
Max staite code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,90 @@
...
 const removeNthElement = (index, array) => {
   // your code here
+  array.splice(index, 1);
 };
 
 const numbersToStrings = numbers => {
   // your code here
+  const numString = numbers.toString();
+  const splitString = numString.split(',');
+  return splitString;
```
StuartCowley, line 35: This same logic could be rewritten as an easily readable a single line:
```
  return numbers.toString().split(',');
```

src/arrays.js
```diff
@@ -1,61 +1,90 @@
 const getNthElement = (index, array) => {
   // your code here
+  return array[index % array.length];
```
StuartCowley, line 3: You should definitely remove the `// your code here` comments fro the entire repo, they are just placeholders and are just cluttering things up now :)

src/arrays.js
```diff
@@ -1,61 +1,90 @@
...
 const removeNthElement2 = (index, array) => {
   // your code here
+  const copy = [...array];
+  copy.splice(index, 1);
+  return copy;
 };
 
 const elementsStartingWithAVowel = strings => {
   // your code here
+  return strings.filter(words => /^[aeiou]/i.test(words));
```
StuartCowley, line 67: Nice regex! Minor point but I would rename the placeholder variable in the filter function to be a singular and not a plural, convention is to use the same word as the plural so in this case it should be 'string':
```
```

src/arrays.js
```diff
@@ -1,61 +1,90 @@
...
 };
 
 const removeSpaces = string => {
   // your code here
+  return string.split(' ').join('');
+
```
StuartCowley, line 73: Dont need the extra blank line here

src/arrays.js
```diff
@@ -1,61 +1,90 @@
...
+  return numbers.reduce((sum, number) => sum + number, 0);
 };
 
 const sortByLastLetter = strings => {
   // your code here
+  return strings.sort((a, b) => {
+    const lastA = a.charAt(a.length - 1);
+    const lastB = b.charAt(b.length - 1);
+    return lastA.localeCompare(lastB);
+  });
```
StuartCowley, line 87: This is good, but you dont actually need the `localeCompare` function here. You can just use the sort function as you are without it, and without the variable creation too:
```
  return strings.sort(
    (a, b) => a.charCodeAt(a.length - 1) - b.charCodeAt(b.length - 1)
  );
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
 function negate(a) {
   // your code here
+  return !(a > 0);
```
StuartCowley, line 3: While this works, it isnt the clearest of logic as the value being passed in is a boolean and not a number. Instead you can just return the inverse of the input like this:
```
  return !a;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 function isSquare(a) {
   // your code here
+  if (a < 0) return false;
+  return Boolean(Math.sqrt(a) % 1 === 0);
 };
 
 function startsWith(char, string) {
   // your code here
+  if (string.startsWith(char)) return true;
+  return false;
```
StuartCowley, line 69: Recursion! This is some thing you rarely see in the wild, and it has very specific use cases so is generally not a good idea to use. Instead, using a simpler comparison would be the way to go:
```
  return char === string[0];
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 function containsVowels(string) {
   // your code here
-};
+  const vowels = ['a', 'e', 'i', 'o', 'u', 'A', 'E', 'I', 'O', 'U'];
+  for (let i = 0; i < string.length; i += 1) {
+    if (vowels.includes(string[i])) {
+      return true;
+    }
+  }
+  return false;
```
StuartCowley, line 80: While this works, its a great example of somewhere that regex would be very useful
This array could be written in the format `/[aeiou]/`

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise.

src/booleans.js
```diff
@@ -1,62 +1,94 @@
 function negate(a) {
   // your code here
+  return !(a > 0);
 };
 
 function both(a, b) {
   // your code here
+  return a > 0 && b > 0;
```
StuartCowley, line 8: As above, you shouldnt try to compare booleans with numerical values like this. Since you comparing 2 booleans it could be written:
```
  return a && b;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function either(a, b) {
   // your code here
+  return a > 0 || b > 0;
```
StuartCowley, line 13: As above:
```
  return a || b;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function none(a, b) {
   // your code here
+  return !(a > 0 || b > 0);
```
StuartCowley, line 18: As above:
```
  return !(a || b);
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function one(a, b) {
   // your code here
+  if (a > 0 && b > 0) {
+    return false;
+  }
+  return a > 0 || b > 0;
```
StuartCowley, line 26: You can simplify this to remove the if statement, and refactor this just slightly:
```
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
+  
+
 
 function truthiness(a) {
-  // your code here
+  return Boolean(a);
```
StuartCowley, line 32: There isnt any use in casting the `a` variable to a Boolean here, as it already is a boolean! Instead, we can use two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function isEqual(a, b) {
   // your code here
+  return Boolean(a === b);
```
StuartCowley, line 37: No need to cast this to a boolean here:
```
  return a === b;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function isGreaterThan(a, b) {
   // your code here
+  return Boolean(a > b);
```
StuartCowley, line 42: As above:
```
  return a > b;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function isLessThanOrEqualTo(a, b) {
   // your code here
+  return Boolean(a <= b);
```
StuartCowley, line 47: And again:
```
  return a <= b;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function isOdd(a) {
   // your code here
+  return Boolean(a % 2);
 };
 
 function isEven(a) {
   // your code here
+  return Boolean(a % 2 == 0);
```
StuartCowley, line 57: The comparison here should use the strict equals operator:
```
  return Boolean(a % 2 === 0);
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 };
 
 function isSquare(a) {
   // your code here
+  if (a < 0) return false;
+  return Boolean(Math.sqrt(a) % 1 === 0);
```
StuartCowley, line 63: You can remove the first line here:
```
  return Math.sqrt(a) % 1 === 0;
```

src/booleans.js
```diff
@@ -1,62 +1,94 @@
...
 
 function isLowerCase(string) {
   // your code here
-};
+  for (let i = 0; i < string.length; i += 1) {
+    if (string[i] !== string[i].toLowerCase()) {
+      return false;
+    }
+  }
+  return true;
```
StuartCowley, line 90: While this works, it can be simplified quire a bit:
```
```

src/numbers.js
```diff
@@ -1,45 +1,60 @@
...
+  return Math.abs(a);
 }
 
 function quotient (a, b) {
   // your code here
+  let quotient2 = Math.floor(a / b);
+  if (a <0 && b >0 || a > 0 && b < 0) {
+     quotient2 += 1;
+  }
+  return quotient2;
```
StuartCowley, line 52: You gave this is a really good go! As it happens theres a built in function for this: Math.trunc (see [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#:~:text=The%20Math.,by%20removing%20any%20fractional%20digits.))
```
  return Math.trunc(a / b);
```

src/objects.js
```diff
@@ -1,41 +1,68 @@
 const createPerson = (name, age) => {
   // your code here
+  const person = new Object();
+  person.name = name, 
+  person.age = age;
+  
+  return person;
```
StuartCowley, line 7: Rather than creating anew object like this and assigning properties, you could instead just return an object directly like this:
```
  return {
    name: name,
    age: age
  };
```
There is a new shorthand way of writing this too, when the keys and values are the same you can just write them singly like this:
```
  return {
    name,
    age
  };
```

src/objects.js
```diff
@@ -1,41 +1,68 @@
...
   // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
   // your code here
+  if (object[property]) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 25: The logic works, but it can be made a bit shorter.

```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,68 @@
...
 };
 
 const isOver65 = person => {
   // your code here
+  if (person.age > 65) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 33: You could simply return the result of evaluating the condition since you are only looking for a boolean value here:
```
  return person.age > 65;
```

src/objects.js
```diff
@@ -1,41 +1,68 @@
...
 };
 
 const getAges = people => {
   // your code here
+  return people.map(persons => persons.age);
```
StuartCowley, line 38: As above, its best to use a singular value of what youre looping over in a map or filter function:
```
```

StuartCowley (commented):
Good stuff! Everything passes, and youve come up with some interesting approaches. 
Ive left some suggestions to refactor your work and make it a bit cleaner in places, plus some shorthand and built in methods that will do some of the work here for you (like with the quotient function)
Also it would be a good idea to update the readme so its not just a set of instructions from the track, but description of what the repo does, how to install and how to test it 

Well done, looks like youre getting a solid start here :)

PR #18, opened 2023-05-20
Armghan z code review

StuartCowley:


src/arrays.js
```diff
@@ -1,61 +1,65 @@
 const getNthElement = (index, array) => {
-  // your code here
+index %= array.length;
+return array[index];
```
StuartCowley, line 3: Did you notice 'Assignment to function parameter' eslint error in your IDE? This is not a good thing to do as it can lead to unexpected behaviours and bugs in larger functions where the value coming into the function is no longer what its expected to be, so you should always avoid doing it.
In this case you can do this using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). This means you dont have to create any variables at all:

```
```

src/arrays.js
```diff
@@ -1,61 +1,65 @@
...
 };
 
 const arrayToCSVString = array => {
-  // your code here
+return array.join();
```
StuartCowley, line 7: This should be indented 2 spaces as its inside a function:
```
  return array.join();
```

src/arrays.js
```diff
@@ -1,61 +1,65 @@
...
 };
 
 const csvStringToArray = string => {
-  // your code here
+  return string.split (',');
 };
 
 const addToArray = (element, array) => {
-  // your code here
+array.push(element);
```
StuartCowley, line 15: This should be indented 2 spaces. The same goes for the rest of the functions in this file, so I would advise going through it and fixing this for all of them rather than me writing each one out again here:
```
  array.push(element);
```

src/arrays.js
```diff
@@ -1,61 +1,65 @@
...
 };
 
 const numbersToStrings = numbers => {
-  // your code here
+return numbers.map(String);
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+ return strings.map(strings => strings.toUpperCase());
```
StuartCowley, line 31: 2 spaces of indentation here not one :) 
Also, pay attention to any errors vscode shows you, often it will explain how to fix it for you. In this case it says "Strings is already declared in the upper scope". This is because youre using the same variable name in the map function which is a bad practice, you should call this something different. 'string' would be a good choice as its not taken and it represents the singular form of what you are looping over, so it reads better too
```
  return strings.map(strings => strings.toUpperCase());
```

src/arrays.js
```diff
@@ -1,61 +1,65 @@
...
-  // your code here
+const array2 = [...array];
+array2.splice(index, 1);
+return array2;
 };
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+regex = /^[aeiou]/gi
+return strings.filter(string => string.match(regex));
```
StuartCowley, line 50: You could put this down into one line:
```
  return strings.filter(string => string.match(/^[aeiou]/gi));
```
Remember when creating a variable you have to use either `let` or `const` too!

src/booleans.js
```diff
@@ -1,61 +1,61 @@
 function negate(a) {
-  // your code here
+  return !a;
 };
 
 function both(a, b) {
-  // your code here
+  return (a&&b);
```
StuartCowley, line 6: The formatting is a bit out here, you should have some spaces around the variables and operators to make this cleaner to read:
```
  return a && b;
```

src/booleans.js
```diff
@@ -1,61 +1,61 @@
...
 };
 
 function either(a, b) {
-  // your code here
+  return (a||b);
```
StuartCowley, line 10: As above
```
  return a || b;
```

src/booleans.js
```diff
@@ -1,61 +1,61 @@
...
 };
 
 function none(a, b) {
-  // your code here
+  return !(a||b);
```
StuartCowley, line 14: As above:
```
  return !(a || b);
```

src/booleans.js
```diff
@@ -1,61 +1,61 @@
...
 };
 
 function one(a, b) {
-  // your code here
+  return (a&&!b) || (!a&&b);
```
StuartCowley, line 18: As above:
```
```
The rest of the function here should be formatted to fit with this style

src/numbers.js
```diff
@@ -1,45 +1,45 @@
 function add (a, b) {
-  // your code here
+    return a+b;
```
StuartCowley, line 2: Just 2 spaces needed:
```
  return a+b;
```

src/numbers.js
```diff
@@ -1,45 +1,45 @@
...
 }
 
 function subtract (a, b) {
-  // your code here
+  return a-b;
```
StuartCowley, line 6: Just spacing for this one and the others below it:
```
  return a - b;
```

src/objects.js
```diff
@@ -1,41 +1,52 @@
 const createPerson = (name, age) => {
-  // your code here
+const person = {
+  name,
+  age
+};
+return person;
```
StuartCowley, line 6: No need to create a variable here, could just return the object as is:
```
  return {
    name,
    age
  };
```

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+return object.hasOwnProperty(property);
```
StuartCowley, line 18: You can fix the eslint error being thrown here by using the array style notation for accessing the properties of the object as follows:
```
  return object[property];
```

```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
+const person = {
+  name,
+  age
+};
+return person;
 };
 
 const getName = object => {
-  // your code here
+return object.name;
```
StuartCowley, line 10: Indentation again here, should be 2 spaces:
```
  return object.name;
```

As before, this should apply to all the other functions here

src/objects.js
```diff
@@ -1,41 +1,52 @@
...
 const averageAge = people => {
-  // your code here
+return people.reduce((acc, person) => acc + person.age, 0) / people.length;
 };
 
 const createTalkingPerson = (name, age) => {
-  // your code here
+  const newPerson = {
+    name: name,
+    age: age,
```
StuartCowley, line 44: You can use the same shorthand you did for `createPerson` here:
```
    name,
    age,
```

StuartCowley (commented):
Good logic here, everything works and tests pass, but there are a lot of formatting issues to sort out and the indentation is off in almost every file. This might not seem that important, but its vital to get this right as its a prerequisite of all devs to be able to write code that follows formatting guidelines. 
I would _highly_ recommend installing prettier to format your code whenever you save it, it would save you a ton of work and keep your code clean without any extra effort on your part

That said once you fix these issues, this repos golden. Keep submitting stuff, if you have any issues setting up formatting tools do ask the tutors and we'll help you out

PR #19, opened 2023-08-11
Lisa wyers code review

StuartCowley:


src/numbers.js
```diff
@@ -1,45 +1,51 @@
...
 }
 
-function quotient (a, b) {
-  // your code here
+function quotient(a, b) {
+  const result = a / b;
+  if (result === -4.5) {
+    return -4;
+  }
+  return Math.floor(result);
```
StuartCowley, line 44: Theres a built in function for this: Math.trunc (see [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#:~:text=The%20Math.,by%20removing%20any%20fractional%20digits.))
```
  return Math.trunc(a / b);
```

src/__tests__/objects.test.js
```diff
@@ -1,4 +1,4 @@
-const {
+qconst {
```
StuartCowley, line 1: This must have been a typo! This q is causing the whole test suite to fail
```
const {
```

src/objects.js
```diff
@@ -1,42 +1,53 @@
 const createPerson = (name, age) => {
-  // your code here
+  return {
+    name: name
+    age: age
```
StuartCowley, line 4: This is missing a comma so is not running:
```
  return {
    name: name,
    age: age
  };
```
There is a new shorthand way of writing this too, when the keys and values are the same you can just write them singly like this:
```
  return {
    name,
    age
  };
```

src/objects.js
```diff
@@ -1,42 +1,53 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object.hasOwnProperty(property);
```
StuartCowley, line 17: You can fix the eslint error being thrown here by using the array style notation for accessing the properties of the object as follows:
```
  return object[property];
```

```
```
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,42 +1,53 @@
...
 
 const isOver65 = person => {
-  // your code here
+  return person.age > 65 ? true : false;
 };
 
 const getAges = people => {
-  // your code here
+  const result = people.map(person => person.age);
+  return result;
```
StuartCowley, line 26: You could just return the result directly here:
```
```

src/objects.js
```diff
@@ -1,42 +1,53 @@
...
-};
+  const totalAge = people.reduce((sum, person) => sum + person.age, 0);
+  return totalAge / people.length;
+};
+
+function createTalkingPerson(name, age) {
+  return {
+    name,
+    age,
+    introduce: function(otherPersonName) {
```
StuartCowley, line 46: You don't need the function keyword here
```
    introduce(otherPersonName) {
```

src/booleans.js
```diff
@@ -1,61 +1,67 @@
...
-function isGreaterThan(a, b) {
-  // your code here
+const isGreaterThan = (a, b) => {
+  return a > b;
 };
 
-function isLessThanOrEqualTo(a, b) {
-  // your code here
+const isLessThanOrEqualTo = (a, b) => {
+  return a >= b;
```
StuartCowley, line 34: You have this backwards:
```
  return a <= b;
```

src/arrays.js
```diff
@@ -1,61 +1,74 @@
...
+  return numbers.filter(number => number % 2 === 0);
 };
 
 const removeNthElement2 = (index, array) => {
-  // your code here
-};
-
-const elementsStartingWithAVowel = strings => {
-  // your code here
+  return array.filter((_, itemIndex) => itemIndex !== index);
```
StuartCowley, line 50: You also deleted the `elementsStartingWithAVowel` function here which is causing the test suite to fail as it cant find the function anymore!

To unblock you you could do this is several ways, one of which is to have an array of values to check for, then use a combination of the filter and includes array methods, and the toLowerCase string method:

```

}
```

src/arrays.js
```diff
@@ -1,61 +1,74 @@
...
 
 const addToArray = (element, array) => {
-  // your code here
+  array.push(element);
 };
 
 const addToArray2 = (element, array) => {
-  // your code here
+  const newArray = [...array, element];
+  return newArray;
```
StuartCowley, line 20: No need for the variable declaration here:
```
  return [...array, element];
```

src/arrays.js
```diff
@@ -1,61 +1,74 @@
...
 };
 
 const arrayToCSVString = array => {
-  // your code here
+  return array.join();
 };
 
 const csvStringToArray = string => {
-  // your code here
+  return string.split();
```
StuartCowley, line 11: For the split method to work you need to pass it a character to split the string around, in this case a comma:
```
  return string.split(',');
```

StuartCowley (commented):
There are a few things to fix here, most of the tests were being skipped but after unskipping them, most of your code works. There are a  few fails that can be easily fixed, and a couple of typo things that will get things running again, Ive added comments on how to fix all these.
In future always make sure your tests are all running before submitting, and if there are any issues you'd like to know more about or don't understand please bring it up in the cohort channel and get a discussion going, everyone gets to help each other out that way :)

PR #20, opened 2023-10-28
Thasnima begum Code review

StuartCowley:


src/arrays.js
```diff
@@ -1,62 +1,70 @@
 const getNthElement = (index, array) => {
-  // your code here
+  while (index>=array.length){
+    index = index - array.length
+  }
+  return array[index];
```
StuartCowley, line 5: This is an interesting approach! Its not good practice to reassign function parameters like this though, as it can lead to confusing code and unwanted side effects. 
An alternative is to use the modulus operator, which I see you've used in the numbers file. Since that will return a remainder instead of a decimal when dividing, it gives a whole number (EG 5 % 3 = 2). 
```
```

src/arrays.js
```diff
@@ -1,62 +1,70 @@
...
 const addToArray = (element, array) => {
-  // your code here
+  array.push(element)
 };
 
 const addToArray2 = (element, array) => {
-  // your code here
+  return array.concat(element)
+ 
+
```
StuartCowley, line 23: Don't need this extra whitespace here

src/arrays.js
```diff
@@ -1,62 +1,70 @@
...
 };
 
 const removeNthElement = (index, array) => {
-  // your code here
+  return array.splice(index, 1)
+
```
StuartCowley, line 28: As above don't need the blank line here

src/arrays.js
```diff
@@ -1,62 +1,70 @@
...
 };
 
 const numbersToStrings = numbers => {
-  // your code here
+  return numbers.join().split(',')
 };
 
 const uppercaseWordsInArray = strings => {
-  // your code here
+  return strings.map((word)=>word.toUpperCase())
```
StuartCowley, line 36: Would be really good to get your IDE set up to automatically format your work on save, as it would make things easier to read and save you some extra typing to boot. In VSCode, if you go into your settings and search for 'format on save', turn this on (in both your workspace settings AND user settings), when you save your file it will automatically format it for you so you don't even have to remember to do it!
This is what this would look like after doing that:
```
  return strings.map(word => word.toUpperCase());
```
Rather than comment similar throughout your work here, Ill just point out logic or other non-formatting type issues as if you were to do this with your project, it will fix all the extra whitespace/not enough spacing/unnecessary syntax for you!

src/arrays.js
```diff
@@ -1,62 +1,70 @@
...
 };
 
 const removeNthElement2 = (index, array) => {
-  // your code here
+  return array.filter(item=>array.indexOf(item)!=index)
 };
 
 const elementsStartingWithAVowel = strings => {
-  // your code here
+  return strings.filter(word=>/^[aeiou]/i.test(word))
```
StuartCowley, line 52: Nice regex!

src/arrays.js
```diff
@@ -1,62 +1,70 @@
...
-  // your code here
+  return numbers.reduce((a,b)=>a+b,0)
 };
 
 const sortByLastLetter = strings => {
-  // your code here
-};
+  
+  strings.sort((a,b)=>a.charCodeAt(a.length-1)-b.charCodeAt(b.length-1))
+  return strings
```
StuartCowley, line 66: You could just return the result of the function directly here:
```
  return strings.sort(
    (a, b) => a.charCodeAt(a.length - 1) - b.charCodeAt(b.length - 1)
  );
```

src/booleans.js
```diff
@@ -1,61 +1,66 @@
...
 };
 
 function none(a, b) {
-  // your code here
+  return !a&&!b
 };
 
 function one(a, b) {
-  // your code here
+  return (a&&!b || !a&b)? true:false;
```
StuartCowley, line 18: This could be simplified as above by just returning the result of the expression:
```
  return (a && b === false) || (b && a === false);
```

src/booleans.js
```diff
@@ -1,61 +1,66 @@
...
 };
 
 function truthiness(a) {
-  // your code here
+  return a? true: false;
```
StuartCowley, line 22: The logic here is fine, but it can be condensed by using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/booleans.js
```diff
@@ -1,61 +1,66 @@
...
 
 function isSquare(a) {
-  // your code here
+  return Number.isInteger(Math.sqrt(a))
+
 };
 
 function startsWith(char, string) {
   // your code here
+  return string.charAt(0) === char
```
StuartCowley, line 52: You should remove the '// your code here' comments

src/booleans.js
```diff
@@ -1,61 +1,66 @@
...
 
 function containsVowels(string) {
   // your code here
+  return /[aeiou]/gi.test(string);
 };
 
 function isLowerCase(string) {
   // your code here
+  return !/[ABDCDEFGHIJKLMNOPQRSTUVWXYZ]/g.test(string);
+
```
StuartCowley, line 63: You could do this without the regex:
```
```

src/objects.js
```diff
@@ -1,41 +1,65 @@
...
 };
 
 const getProperty = (property, object) => {
-  // your code here
+  return object[property];
 };
 
 const hasProperty = (property, object) => {
-  // your code here
+  return object[property]? true: false;
```
StuartCowley, line 14: While this works, we could use that double negation trick to reduce this down aa bit:
```
  return !!object[property];
```
Much cleaner :)

src/objects.js
```diff
@@ -1,41 +1,65 @@
...
 };
 
 const isOver65 = person => {
-  // your code here
+  return person.age>65? true:false;
```
StuartCowley, line 18: You don't need the ternary here, you can just return the result of the expression:
```
  return person.age > 65;
```

src/objects.js
```diff
@@ -1,41 +1,65 @@
...
+  return array
 };
 
 const findByName = (name, people) => {
-  // your code here
+for (let i=0;i<people.length;i++){
+  if (people[i].name === name){
+    return people[i]
+  }
+}
```
StuartCowley, line 34: You could use the built in filter method here for this:
```
```

src/objects.js
```diff
@@ -1,41 +1,65 @@
...
 
 const findHondas = cars => {
-  // your code here
+  const array = []
+  for (let i=0;i<cars.length;i++){
+     if (cars[i].manufacturer==="Honda"){
+      array.push(cars[i])
+     }
+  }
+  return array
```
StuartCowley, line 44: As above, the filter method can do this for us:
```
```

StuartCowley (commented):
Good work! Everything passes, though some things could be done in a cleaner or more efficient way, plus the formatting could use some attention. Ive put suggestions throughout to sort these things out, as well as how to set up your IDE to handle formatting things for you automatically on save

The logic itself is sound though, well done!

